PR #1, opened 2011-12-13
Add "increment/decrement by N" to pncounter

russelldb:
Create a G-Counter
Replace use of core vclock with G-Counter
Add tests for G-Counter


seancribbs:
This looks good, it also removes the dependency on riak_core's vclock impl. +1 to merge

PR #2, opened 2011-12-15
Persist in bitcask, not dets

russelldb:
refactor db stuff into simple internal API for easy swapping of storage

### DIFF
--- a/apps/riak_crdt/src/riak_crdt_vnode.erl
+++ b/apps/riak_crdt/src/riak_crdt_vnode.erl
@@ -189,7 +189,7 @@ store({Key, Value}, StorageState) ->
 fold(Fun, Acc, StorageState) ->
     bitcask:fold(StorageState, Fun, Acc).
 
--spec is_empty(StorageState :: term()) ->
+-spec db_is_empty(StorageState :: term()) ->
                       boolean().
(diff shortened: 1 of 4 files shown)


apps/riak_crdt/src/riak_crdt_vnode.erl
```diff
@@ -137,5 +124,82 @@ handle_coverage(_Req, _KeySpaces, _Sender, State) ->
...
+store({Key, Value}, StorageState) ->
+    MKey = make_mkey(Key),
+    bitcask:put(StorageState, MKey, term_to_binary(Value)).
+
+-spec fold(function(), list(), StorageState :: term()) ->
+                  list().
+fold(Fun, Acc, StorageState) ->
+    bitcask:fold(StorageState, Fun, Acc).
+
+-spec is_empty(StorageState :: term()) ->
```
seancribbs: `db_is_empty` is the function name.

seancribbs:
We should probably dialyze this before merging (after fixing that one -spec line).

russelldb:
It has been dialyzed. There is a the spec fix and a couple more dialyzer fixes. However, I'd already started a dialyzer branch off the API branch, so there is one fix in their too.

Issue #20, opened 2012-09-06
Restructure files to fit OTP single-app structure


seancribbs:
This will be eventually embedded in the `riak` release chain, so it doesn't need to be a standalone embedded node.

Issue #15, opened 2012-09-06
Rename project to riak_dt


seancribbs:
Also rename files/module names.

PR #23, opened 2012-09-07
Bump riak_core version to latest

russelldb:
Fix some whitespace


seancribbs:
Needs more `-ifdef(EQC).`

russelldb:
OK

seancribbs:
Looks like this was applied to #20 / #15, so closing.

PR #22, opened 2012-09-08
Add licence to head of all files

russelldb:
Add licence to head of all files before we open source


src/riak_dt_wm_orset.erl
```diff
@@ -1,3 +1,25 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_dt_wm_orset: Webmachine resource ofor OR-Set access
```
seancribbs: typo `ofor`

seancribbs:
Similar to the other repos, we should also have in the root a boilerplate LICENSE file which is a copy of the Apache 2.0 license.

seancribbs:
:+1: good to merge

PR #31, opened 2012-09-10
Add README to riak_dt

russelldb:
If we are going to open source we need at least a cursory README in the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,14 +3,14 @@
 ## WHAT?
 
 Currently under initial development, riak_dt is a platform for
-convergent data types. Built on
+convergent data types. It's built on
 [riak core](https://github.com/basho/riak_core) and deployed with
-[riak](https://github.com/basho/riak). All our current work is around
-supporting fast, replicated, eventually consistent counters (though
-more data types are in the repo, and on the way.) This work is based
-on the paper
-["A Comprehensive study of Convergent and Commutative Replicated Data Types"]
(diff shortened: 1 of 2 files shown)


PharkMillups:
Good stuff. A few nit-picky grammar/markup things:

> Built on riak core and deployed with riak.

It's built on riak core and deployed with riak. 

> All our current work is around supporting fast, replicated, eventually consistent counters (though more data types are in the repo, and on the way.)

All of our current work is around supporting fast, replicated, eventually consistent counters (though more data types are in the repo, and on the way).  

> This work is based on the paper "A Comprehensive study of Convergent and Commutative Replicated Data Types". which you may find an interesting read.

This work is based on the paper - A Comprehensive study of Convergent and Commutative Replicated Data Type -  which you may find an interesting read.

> Which should have fetched all the riak dependancies from github and compiled them, and create 4 release directories under the dev directory. Now start your nodes and build your cluster.

Ths should have fetched all the riak dependancies from github, compiled them, and created four release directories under the `dev` directory. 

Now start your nodes and build your cluster.

> There is no explicit create operation for a counter

There is no explicit `create` operation for a counter

> You can decrement a counter too:

You can decrement a counter, too:

russelldb:
Added a commit which incorporates your suggestions. Please to be re-reviewing.

seancribbs:
:+1: assuming @PharkMillups likes it.

PharkMillups:
Phark approves. Phark will permit merge.

PR #35, opened 2012-09-11
Use a binary uuid for actor id

russelldb:
I tried with both druuid and uuid and in my (micro) benchmark the pure erlang uuid.erl module was faster, so I went with that. _HOWEVER_ we'd better check about the legal status of that licence header from Travis Vachon. 

Speed is probably not _really_ an issue, since at this point we are only talking about uuids for vnodeids. But, if we start using them for U-Sets then maybe speed matters.

Let's talk.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -3,7 +3,7 @@
 {cover_enabled, true}.
 {erl_opts, [debug_info, warnings_as_errors, {parse_transform, lager_transform}]}.
 {edoc_opts, [{dir, "../../doc"}]}.
-{deps, [{riak_core, "1.2.*",
-         {git, "git://github.com/basho/riak_core", "master"}},
+{deps, [
+        {riak_core, "1.2.*", {git, "git://github.com/basho/riak_core", "master"}},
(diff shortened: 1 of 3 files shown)


seancribbs:
That license reads like the MIT or BSD license. Note that the `-import(random).` directive has no effect since it doesn't import any functions.

seancribbs:
Can you share your benchmarks? I'd love to see the comparison.

russelldb:
Heh, 'benchmarks' is timer:tc a few thousand times. But given the infrequency of use, maybe druuid makes more sense for the license alone.

seancribbs:
@russelldb Yes, this looks to be http://opensource.org/licenses/BSD-3-Clause. I'm ok with inclusion as long as no other deps have a uuid generator that we can reuse. Also, as you say, setting it on startup is relatively infrequent and not in the path of normal operations.

The rest of this is so simple, :+1: to merge.

russelldb:
Going to merge it, and add concern about license as an issue. Maybe just check with Dizzy/Jon/Justin at a later date before we ship/flip the bit.

PR #13, opened 2012-09-11
Make storage backends pluggable

seancribbs:
We should be able to reuse Riak KV storage backends or make them otherwise pluggable.

### DIFF
--- a/src/riak_dt_vnode.erl
+++ b/src/riak_dt_vnode.erl
@@ -227,6 +227,10 @@ do_merge({Mod, Key}, RemoteVal, StorageState) ->
     end.
 
 
+%% -------------------
+%% Storage engine
+%% -------------------
+
 %% @doc Initializes the internal storage engine where datatypes are
 %% persisted.
 -spec start_storage(Partition :: integer()) ->
@@ -239,7 +243,6 @@ start_storage(Partition) ->
         Cask ->
             {ok, Cask}
     end.
-
 
 %% @doc Stops the internal storage engine.
 -spec stop_storage(StorageState :: term()) ->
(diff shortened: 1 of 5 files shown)


seancribbs:
So, I decided to back-off of the "pluggable backends" idea in favor of just making sure that our vnode code is well documented, -spec'd and includes the capacity to merge the bitcasks. I suspect our usage will be creating lots of garbage, so we might at a later time want to segregate our bitcask options into a different config so that merging can be tuned. I'm happy to continue down that road if desired, but there are other issues more pressing.

src/riak_dt_vnode.erl
```diff
@@ -170,23 +237,34 @@ do_merge({Mod, Key}, RemoteVal, StorageState) ->
...
+
+%% @doc Initializes the internal storage engine where datatypes are
+%% persisted.
+-spec start_storage(Partition :: integer(), Options::list()) ->
                            StorageState :: term().
-start_storage(Partition) ->
+start_storage(Partition, Options0) ->
     {ok, PartitionRoot} = get_data_dir(Partition),
-    case bitcask:open(PartitionRoot, [read_write]) of
+    Options = set_mode(read_write, Options0),
```
russelldb, line 251: I don't understand the purpose of this function call.

russelldb:
Apart from where I don't get the call to `set_mode/2` it all looks good to me. Passes tests and basic cluster start, increment, decrement, partition, heal smoke test. +1 (after I understand `set_mode/2`

seancribbs:
@russelldb `set_mode/2` just ensures that the `read_write` flag is there and the `read_only` flag is not, within the options list.

russelldb:
@seancribbs  Ok, but, why have two heads then? I can't see how `set_mode(read_only, Options0)` would ever be called.

seancribbs:
@russelldb Sorry, that was copy-pasted from riak_kv_bitcask_backend.

PR #16, opened 2012-09-12
Modify update FSM to require updates going to primaries

seancribbs:
Fallbacks should not receive update operations, but only merge-style replication for the sake of durability and availability (essentially PW=1). We can work this into the FSM as a preparation phase that ensures at least one primary is available and that it is selected for updates.

### DIFF
--- a/src/riak_dt_vnode.erl
+++ b/src/riak_dt_vnode.erl
@@ -331,7 +331,7 @@ make_mkey(ModKey) ->
 get_data_dir(Partition) ->
     DataRoot = app_helper:get_env(riak_dt, data_root, "data/riak_dt_bitcask"),
     PartitionRoot = filename:join(DataRoot, integer_to_list(Partition)),
-    ok = filelib:ensure_dir(PartitionRoot),
+    ok = filelib:ensure_dir(filename:join(PartitionRoot, ".dummy")),
     {ok, PartitionRoot}.
 
 
--- a/src/riak_dt.app.src
+++ b/src/riak_dt.app.src
@@ -8,6 +8,7 @@
                   kernel,
                   stdlib,
                   lager,
+                  bitcask,
                   webmachine,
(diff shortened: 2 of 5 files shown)


src/riak_dt_vnode.erl
```diff
@@ -331,7 +331,7 @@ make_mkey(ModKey) ->
 get_data_dir(Partition) ->
     DataRoot = app_helper:get_env(riak_dt, data_root, "data/riak_dt_bitcask"),
     PartitionRoot = filename:join(DataRoot, integer_to_list(Partition)),
-    ok = filelib:ensure_dir(PartitionRoot),
+    ok = filelib:ensure_dir(filename:join(PartitionRoot, ".dummy")),
```
russelldb, line 334: I don't understand the purpose of the 'dummy' suffix. Oh wait, now I do after I RTFM. 

```
The ensure_dir/1 function ensures that all parent directories for the given file or directory name Name exist,
trying to create them if necessary.
```

_Parent_ being the key.

russelldb:
Test fails for me in setup. Looks like `riak_core:wait_for_service(riak_dt).` waits and waits and finally exits with `badarg`. Looks like the ets table node watcher uses goes away. Digging into it but not getting anywhere yet.

Works fine if I start up a Riak node with riak_dt, though.

russelldb:
Rebar isn't comiling my bitcask nifs. Can you give me an example of how you ran this test since we no longer have rebar in the riak_dt repo?

russelldb:
Ok, finally, with a new rebar I got it to pass. Sorry for all the noise, thanks @seancribbs +1 to merge.

src/riak_dt_vnode.erl
seancribbs, line 334: Right, so it doesn't create `.dummy` but all directories above it.

PR #38, opened 2012-09-14
Fix crasher in `delete` callback from badmatch of start_storage return

russelldb:
Missed this in review of sdc-storage-backends


seancribbs:
:+1:

PR #30, opened 2012-09-17
Drop riak_dt_console

seancribbs:
Drop separate console commands.
Consider if any console commands are needed specifically for rdt


russelldb:
WFM +1

PR #40, opened 2012-09-25
Seed random for uuid generation

russelldb:
If 'random' isn't seeded then duplicate vnodeids happen, and that's bad, since actors update each others entry in the counter.


russelldb:
riak_test caught this. Yay, riak_test!

seancribbs:
:+1: Simple change, makes sense. If we want to get better PRNGs later we could change uuid to use crypto:rand_uniform() instead.

PR #32, opened 2012-10-04
Fix handoff command behavior

seancribbs:
While in the handoff state, the `riak_dt_vnode` only handles fold requests, all others will cause a `function_clause` exception. We should figure out what the behavior should be during handoff.


src/riak_dt_vnode.erl
```diff
@@ -287,10 +359,27 @@ store({Key, Value}, StorageState) ->
     bitcask:put(StorageState, MKey, term_to_binary(Value)).
 
 %% @doc Folds over the persistent storage.
--spec fold(function(), list(), StorageState :: term()) ->
-                  list().
-fold(Fun, Acc, StorageState) ->
-    bitcask:fold(StorageState, Fun, Acc).
+%% -spec fold(function(), list(), StorageState :: term()) ->
```
russelldb: Do you need to leave the old, commented out, code here?

seancribbs: Nope, sorry.

russelldb:
Tests pass, comments fantastic. Thank you! +1

PR #33, opened 2012-10-05
Ignore unsupported coverage requests

seancribbs:
Currently a coverage request will cause the vnode to stop/exit. Some coverage requests might need implementing (list-keys?) but others should just be ignored rather than killing the vnode.


russelldb:
+1

PR #39, opened 2012-10-10
Fix semantics of counters resource

seancribbs:
The current counters increment and decrement like so:

```
POST /counters/a/increment
POST /counters/a/decrement
```

This is a bad practice as it puts the verb in the URL. It would be better to use form parameters or JSON to submit the "change". curl examples:

``` bash
curl http://riakhost:8098/counters/a -F increment=5
curl http://riakhost:8098/counters/a -F decrement=2
curl http://riakhost:8098/counters/a -d '{"increment":5}' -H "Content-Type: application/json"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ that doesn't exist it will be created for you.
 
 Create / increment a counter named `a`:
 
-    curl -X POST localhost:8091/counters/a/increment
+    curl -X POST localhost:8091/counters/a -d'1'
 
 Read it from any and all nodes:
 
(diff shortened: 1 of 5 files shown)


russelldb:
Is it possible that the README could be updated to reflect this change, too, please?

seancribbs:
Will do.

Sean Cribbs

seancribbs:
@russelldb Fixed the examples. Travis build is failing, but could you run the WM tests locally?

russelldb:
Hmm, test times out for me locally. Will dig a little.

seancribbs:
@russelldb This may be similar to the timeouts I would occasionally see testing riak_api. I'll find out what I did there and possibly include an additional wait.

seancribbs:
@russelldb I found the culprit, the vnode test wasn't calling `meck:unload()` at the end of a test with mocks installed. This broke the vnode, making it impossible to send a request to any of them. Now it works and plows through the tests quickly.

russelldb:
Works for me. Many thanks +1.

PR #45, opened 2013-07-15
Garbage Compaction Ideas - Do Not Merge

lenary:
Just a central place to throw comments at these ideas about garbage compaction.

Flow is expected to be:
1. Work out Thresholds etc
3. Acquire consensus (not included in this code), including making a new "epoch" somehow and electing leader
4. Leader calls `gc_propose/2` to get some kind of object that encapsulates what should be done to the CRDT
5. This object is broadcast to all replicas and they all (including the leader) call `gc_execute/2` with that object and the CRDT to actually compact.
6. Consensus is released.

Upon merge of a replica that has fallen behind with its compaction, it will first do any pending Compactions (got from the replica its trying to merge with) it hasn't already done, and then merge.

### DIFF
--- a/include/riak_dt_gc.hrl
+++ b/include/riak_dt_gc.hrl
@@ -6,6 +6,7 @@
 -define(GC_META_ACTOR(GM), GM#riak_dt_gc_meta.actor).
 -record(riak_dt_gc_meta,
         {
+          epoch :: epoch(),             % Epoch of GC
           actor :: actor(),             % Actor performing the GC
           primary_actors :: [actor()],  % Actors most likely to be involved in operations
           readonly_actors :: [actor()], % Actors that can't be GCd (ie cluster remotes)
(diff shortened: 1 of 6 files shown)


lenary:
Thinking about the process, despite `gc_ready/2` having said GC should happen, should we perhaps have a way for `gc_propose/2` to say "actually don't do this GC"? It would just call into `gc_ready/2` and return false or something, in case merges between calling `gc_ready/2` and aquiring the leadership have made the crdt not worth GCing.

Maybe we should just call `gc_ready/2` inside the consensus again before propose/execute, or maybe we should still GC, as we've done the expensive part (consensus) already.

So, 3 options:
- return something special from gc_propose/2 if gc is not worth doing (despite result of gc_ready/2)
- call gc_ready/2 again after grabbing consensus to make sure
- GC anyway

I assume the answer to this will become clearer after some more implementation.

lenary:
Moving all work into #46, (essentially starting again, remembering the ideas from this PR).

lenary:
Deleting this branch, I think we have everything we need from it @seancribbs

PR #46, opened 2013-07-27
Fragment-based GC [do not merge]

lenary:
Taking all the good ideas from #45 and essentially starting again.

Flow is expected to be:
1. Work out Thresholds etc, for the meta-object (this includes making a new epoch, maybe)
3. Acquire consensus (not included in this code), including electing leader
4. Leader calls `gc_get_fragment/2` to get the fragment of the CRDT that should be removed (for saving somewhere else)
5. This fragment is broadcast to all replicas and they all (including the leader) call `gc_replace_fragment/3` with that object and the CRDT. This is the compaction step.
6. Consensus is released.

The meta object contains data on primary actors, readonly actors (cluster remote views), epoch, and a threshold of how many actors are allowed compared to the number of primary+readonly actors. 

I'm going to close #45, but the ideas are still in my head.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ dev/*
 doc/*
 rel/riak_dt
 data/*
+.eqc-info
+current_counterexample.eqc
--- a/include/riak_dt_gc_meta.hrl
+++ b/include/riak_dt_gc_meta.hrl
@@ -1,0 +1,15 @@
+
+-type actor() :: term().
+-type epoch() :: {actor(), erlang:timestamp()}.
+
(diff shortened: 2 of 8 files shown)


lenary:
Hopefully b73b99a should make it clear how `compact_threshold` is used. I added a macro that should be helpful. There are a particular set of semantics associated with it (documented in `include/riak_dt_gc_meta.hrl`)

Most `gc_ready/2` functions I expect to use it, but they can also use other conditions not related to `compact_threshold` if it makes sense.

PR #47, opened 2013-07-27
Optimised PN-Counter Format (aka PNCounters V2)

lenary:
The first version of the PNCounter code literally used a pair of GCounters as its representation and deferred a lot of the behaviour down to GCounter functions.

However, there were issues with this, essentially to do with keeping the two lists of actors in sync (something needed for GC) and the fact that you store each actor twice, which is inefficient.

The new format is very similar to the structure of a GCounter, but the actor-count tuples include a count of both increments and decrements. For example:

```
Old Style: {[{a,4}],[{a,1},{b,2}]}
New Style: [{a,4,1},{b,0,2}]
```

I've updated `from_binary/1` to convert into the new structure every time, however there's also a new `to_binary/2` which can create the old style (should it be required for downgrades).

### DIFF
--- a/src/riak_dt_pncounter.erl
+++ b/src/riak_dt_pncounter.erl
@@ -144,9 +144,7 @@ to_binary(PNCnt) ->
 -spec to_binary(pncounter() | v1_pncounter(), v2 | v1) -> binary().
 to_binary(PNCnt, v2) ->
     PNBin = term_to_binary(PNCnt),
-    PNSize = byte_size(PNBin),
-    <<?TAG:8/integer, ?V2_VERS:8/integer, 
-      PNSize:32/integer, PNBin/binary>>;
+    <<?TAG:8/integer, ?V2_VERS:8/integer, PNBin/binary>>;
 to_binary({P,N}, v1) ->
     PBin = riak_dt_gcounter:to_binary(P),
     NBin = riak_dt_gcounter:to_binary(N),
@@ -158,8 +156,7 @@ to_binary({P,N}, v1) ->
 
 %% @doc Decode a binary encoded PN-Counter
 -spec from_binary(binary()) -> pncounter().
-from_binary(<<?TAG:8/integer, ?V2_VERS:8/integer,
-              PNBinLen:32/integer, PNBin:PNBinLen/binary>>) ->
+from_binary(<<?TAG:8/integer, ?V2_VERS:8/integer, PNBin/binary>>) ->
     binary_to_term(PNBin);
 from_binary(<<?TAG:8/integer, ?V1_VERS:8/integer,
               PBinLen:32/integer, PBin:PBinLen/binary,
--- a/src/riak_dt_pncounter.erl
+++ b/src/riak_dt_pncounter.erl
@@ -36,7 +36,7 @@
 
 -export([new/0, new/2, value/1, value/2,
          update/3, merge/2, equal/2, to_binary/1, from_binary/1]).
--export([to_binary/2, from_binary/2, change_versions/3]).
+-export([to_binary/2, from_binary/2, current_version/1, change_versions/3]).
 
 %% EQC API
 -ifdef(EQC).
--- a/src/riak_dt_pncounter.erl
+++ b/src/riak_dt_pncounter.erl
@@ -35,7 +35,8 @@
 -module(riak_dt_pncounter).
 
 -export([new/0, new/2, value/1, value/2,
-         update/3, merge/2, equal/2, to_binary/1, to_binary/2, from_binary/1]).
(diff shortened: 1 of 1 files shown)


src/riak_dt_pncounter.erl
```diff
@@ -89,36 +91,60 @@ value(negative, {_P, N}) ->
...
 -spec update(pncounter_op(), term(), pncounter()) -> pncounter().
-update(increment, Actor, {Incr, Decr}) ->
-    {riak_dt_gcounter:update(increment, Actor, Incr), Decr};
-update({increment, By}, Actor, {Incr, Decr}) when is_integer(By), By > 0 ->
-    {riak_dt_gcounter:update({increment, By}, Actor, Incr), Decr};
-update(decrement, Actor, {Incr, Decr}) ->
-    {Incr, riak_dt_gcounter:update(increment, Actor, Decr)};
-update({decrement, By}, Actor, {Incr, Decr}) when is_integer(By), By > 0 ->
-    {Incr, riak_dt_gcounter:update({increment, By}, Actor, Decr)}.
+update(increment, Actor, PNCnt) ->
```
russelldb, line 97: I regret this `{increment, By} / {decrement, By}` interface. Do you think we're stuck with it now? I wish I'd just done `{incr, Amt}` and allowed negative values, and even `0` (just return current.)

src/riak_dt_pncounter.erl
```diff
@@ -89,36 +91,60 @@ value(negative, {_P, N}) ->
...
 
 %% @doc Merge two `pncounter()'s to a single `pncounter()'. This is the Least Upper Bound
 %% function described in the literature.
 -spec merge(pncounter(), pncounter()) -> pncounter().
-merge({Incr1, Decr1}, {Incr2, Decr2}) ->
-    MergedIncr =  riak_dt_gcounter:merge(Incr1, Incr2),
-    MergedDecr =  riak_dt_gcounter:merge(Decr1, Decr2),
-    {MergedIncr, MergedDecr}.
+merge(PNCntA, PNCntB) ->
+    % TODO
```
russelldb: Looks @TODONE to me.

lenary: Sorted in f66392f

src/riak_dt_pncounter.erl
lenary, line 97: In f66392f I have added clauses that allow you to supply `{increment, Negative}` (falls through to `{decrement, -Negative}`), and `{increment | decrement, 0}` (a no-op)

src/riak_dt_pncounter.erl
```diff
@@ -89,36 +91,63 @@ value(negative, {_P, N}) ->
...
 %% @doc Encode an effecient binary representation of `pncounter()'
 -spec to_binary(pncounter()) -> binary().
-to_binary({P, N}) ->
+to_binary(PNCnt) ->
+    to_binary(PNCnt, v2).
+
+-spec to_binary(pncounter() | v1_pncounter(), v2 | v1) -> binary().
+to_binary(PNCnt, v2) ->
+    PNBin = term_to_binary(PNCnt),
+    PNSize = byte_size(PNBin),
```
russelldb: Why the size? Can't we just decode by matching on Tag and Version?

russelldb:
Apart from the redundant size field in the v2 binary, this looks good to me. Have you micro benchmarked for size vs. speed against the original implementation?

lenary:
I have not yet microbenchmarked for speed. I have, however, microbenchmarked for size using this code: https://gist.github.com/lenary/e3f583045d3955cbb846

Results (percentage = v2 is x% the size of v1, Total operations: 10000)

```
   2 Actors: 54.6%
   3 Actors: 56.7%
   4 Actors: 57.9%
   5 Actors: 58.7%
  10 Actors: 60.5%
 100 Actors: 62.3%
1000 Actors: 57.1%
```

Results (percentage = v2 is x% the size of v1, Total operations: 100000)

```
   3 Actors: 56.8%
   5 Actors: 58.8%
  10 Actors: 60.5%
1000 Actors: 62.5%
```

This doesn't account for any more than the `riak_dt_pncounter:to_binary/2` function (ie metadata and riak object stuff is just meh)

@seancribbs has mentioned capabilities etc, which will be a problem, so I'll do a bit of work on it this week, but will cope with most of those issues later

lenary:
So, as per guidance from @seancribbs, i have coded _every_ conversion function, so we have a `to_binary/2`, `from_binary/2`, `change_versions/3` and `current_version/1`. Next up is making a conversion script, but with all those functions this should be simple.

I also switched to using the constants instead of atoms like `v1` and `v2`, so there's less crazy.

seancribbs:
This looks good, @lenary. Merging!

PR #49, opened 2013-08-07
ORSet V2

lenary:
Gave it a bit of attention to make it more concise and it should also have a smaller size when serialised. :)

EDIT: It seems work has been done on VVORSet, so not touching that.

### DIFF
--- a/src/orset_bm.erl
+++ b/src/orset_bm.erl
@@ -21,7 +21,7 @@
 start(Cmds, Mod) ->
     TS = erlang:now(),
     Coord =  start_coordinator(?N, Mod, self()),
-    [start_proc(I, Cmds, Mod, Coord) || I <- lists:seq(1, ?N)],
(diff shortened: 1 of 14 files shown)


lenary:
As we've not yet deployed any OR-Set implementations (unless I'm wrong about this) I don't think upgrade/downgrade functions are needed.

russelldb:
You are not wrong, no OR-Set in the wild.

lenary:
Ah yes, I'm now reminded that @russelldb did some work on benchmarking this OR-Set. The bm was slightly biased towards a test that vvorset would do better at, so there wasn't much of an improvement, however, i think this OR-Set will do better when there are lots of keys and few adds/removes per key.

russelldb:
Yeah, you could add some less biased scenarios to orset-bm.

seancribbs:
Be sure you include all the callbacks.

PR #51, opened 2013-08-15
New OR-Set implementation, should save muchos bytos

lenary:
Github got confused and closed #49, so work is here (we were screwing around with branches)

Needed: 
- [ ] Benchmark.

### DIFF
--- a/src/riak_dt_orset.erl
+++ b/src/riak_dt_orset.erl
@@ -42,7 +42,7 @@
 -endif.
 
 -export_type([orset/0, binary_orset/0, orset_op/0]).
--opaque orset() :: orddict:orddict(member(), uniques()).
+-opaque orset() :: orddict:orddict().
 
 -type binary_orset() :: binary(). %% A binary that from_binary/1 will operate on.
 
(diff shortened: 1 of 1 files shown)


lenary:
I've rebased onto `develop` and added stuff about precondition errors and context. The model still passes.

I haven't run dialyzer. It probably hates me.

src/riak_dt_orset.erl
```diff
@@ -42,7 +42,7 @@
 -endif.
 
 -export_type([orset/0, binary_orset/0, orset_op/0]).
--opaque orset() :: {entries(), entries()}.
+-opaque orset() :: orddict:orddict(member(), uniques()).
```
seancribbs: `orddict` is a 0-arity type:

```
orddict() = [{Key :: term(), Value :: term()}]
```

src/riak_dt_orset.erl
```diff
@@ -189,8 +122,13 @@ reset(ORSet, Actor) ->
 %% an operation is needed at a replica without sending the entire
 %% state to the client.
 -spec precondition_context(orset()) -> orset().
-precondition_context({ADict, _RDict}) ->
-    {ADict, orddict:new()}.
+precondition_context(ORDict) ->
+    ordict:fold(fun(Elem, Tokens, ORDict1) ->
```
seancribbs: Typo, `ordict` -> `orddict`

seancribbs:
I've run the `orset_bm` against this, it uses way more space than the ORSWOT (proportional to the number of adds), but is not significantly slower for it. :+1:, I'll handle the merge.

lenary:
:hammer: :hand:

PR #52, opened 2013-08-15
Fragment-based GC [do not merge]

lenary:
Github also closed #46. :-1: Well, here it is again. 

Taking all the good ideas from #45 and essentially starting again.

Flow is expected to be:
1. Work out Thresholds etc, for the meta-object (this includes making a new epoch, maybe)
3. Acquire consensus (not included in this code), including electing leader
4. Leader calls `gc_get_fragment/2` to get the fragment of the CRDT that should be removed (for saving somewhere else)
5. This fragment is broadcast to all replicas and they all (including the leader) call `gc_replace_fragment/3` with that object and the CRDT. This is the compaction step.
6. Consensus is released.

The meta object contains data on primary actors, readonly actors (cluster remote views), epoch, and a threshold of how many actors are allowed compared to the number of primary+readonly actors. 

I'm going to close #45, but the ideas are still in my head.


lenary:
I haven't quite worked out how to solve the problems associated with epochs stored in CRDTs that are so simple that they don't support GC (flags, lwwregs). My feeling is that they should not have an epoch field, and instead return a token which effectively says "yo dude, we'll always be 'current'". I don't really know though.

seancribbs:
I don't know that this branch will work as-is, but let's revisit in 2.1.

lenary:
It won't work as-is, and @russelldb wrote his own eqc. maybe we should just abandon this.

lenary:
wait, wrong one, yeah, i think we can just close and ignore this completely.

seancribbs:
Let's leave open, I think we should review this approach for ideas before we try to address GC on a greater scale.

PR #53, opened 2013-08-26
Add preconditions failures to orset / map

russelldb:
Update all modules `update/3` to return `{ok, CRDT} | {error, Fail}`

Also discovered broken EQC tests, which this PR has fixes for, and then assorted fixes for bugs found.

These tests then found errors in the precondition code, which I fixed (multi-ops were not failing atomically.)

There is also a change to the Map to keep tombstones (a bug was dropping them when to replicas that agreed on removal merged, even if a third had a concurrent update.) This leads to some surprising behaviour if two concurrent removes occur concurrent with an update, the reset occurs twice, and for a counter that has partitioned since a merge of all increments, that will result in a double decrement. I imagine it is a very rare edge case, but I note it here as we don't have a n answer yet.

Also changed the reset strategy on LWW to reset the timestamp to zero. This is non-monotonic. It deals with the case where a reset / remove | update causally, but happens temporally after the update, resulting in a present register field with an `undefined` value.

Also, removed `zorset` and `rwmap`, no time to get them correct for 2.0.

Added bulk operations to Set types `(add_all, remove_all, {update, [{add, _} | {remove, _} | etc}`


src/riak_dt_lwwreg.erl
```diff
@@ -109,8 +114,8 @@ equal(_, _) ->
 
 %% @Doc reset to default value
 -spec reset(lwwreg(), term()) -> lwwreg().
-reset({_Value, TS}, _Actor) ->
-    {undefined, TS+1}.
+reset({_Value, _TS}, _Actor) ->
+    {undefined, 0}.
```
lenary, line 118: I wasn't sure why you took it back to zero, but thinking about concurrency in maps I now see why. :D

src/riak_dt_multi.erl
```diff
@@ -63,29 +86,35 @@
...
 -type valuelist() :: [{field(), crdt()}].
 
--type crdt()  :: riak_dt_gcounter:gcounter() | riak_dt_pncounter:pncounter() |
-                 riak_dt_lwwreg:lwwreg() | riak_dt_orset:orset() |
-                 riak_dt_vvorset:vvorset() | riak_dt_zorset:zorset() |
-                 riak_dt_multi:multi().
+-type crdt()  ::  riak_dt_pncounter:pncounter() |
+                  riak_dt_lwwreg:lwwreg() |
+                  riak_dt_vvorset:vvorset() | riak_dt_orset:orset() |
+                  riak_dt_multi:multi().
```
lenary, line 92: You're missing `riak_dt_od_flag` from here, which you have in the `crdt_op()` spec below.

src/riak_dt_multi.erl
```diff
@@ -63,29 +86,35 @@
...
                    riak_dt_lwwreg:lwwreg_op() | riak_dt_orset:orset_op() |
-                   riak_dt_vvorset:vvorset_op() | riak_dt_zorset:zorset_op() |
+                   riak_dt_vvorset:vvorset_op() | riak_dt_od_flag:od_flag_op() |
                    riak_dt_multi:multi_op().
 
 -type multi_q() :: size | {get, field()} | {get_crdt, field()} |
                    keyset | {contains, field()}.
 
+-type values() :: [value()].
+-type value() :: {field(), riak_dt_multi:value() | integer() | [term()] | boolean() | term()}.
```
lenary: Shouldn't this be `riak_dt_multi:values()` instead of `riak_dt_multi:value()` ?

src/riak_dt_multi.erl
```diff
@@ -379,31 +376,82 @@ eqc_state_value({_Cnt, Dict}) ->
...
-                dict:new(),
-                sets:to_list(Remaining)),
-    proplists:get_value(K, dict:to_list(Res), T:new()).
+                              %% if key is in Acc merge with it and replace
+                              dict:update(Key, fun(V) ->
+                                                       Type:merge(V, Value) end,
+                                          Value, Acc) end,
+                      dict:new(),
+                      sets:to_list(Remaining)),
+    case proplists:get_value(K, dict:to_list(Res)) of
```
lenary: What's going on here? Can't you just use the functions in `dict`?

russelldb: What? Feel free to refactor that for me, it was a quick fix for failing eqc.

lenary:
Maybe it's not supposed to be part of this PR, but `riak_dt_orset` has no `reset` function.

src/riak_dt_multi.erl
russelldb: Adding the preconditions fundamentally changed the model, but there was no time for a re-write…hence that grab bag.

russelldb:
No, it needs one _if_ we allow people to store it in Maps. We haven't made the call yet, but right now we are using vvorset.

lenary:
Right, my only note is that https://github.com/basho/riak_dt/pull/53/files#L5R91 suggests it does, however, I get that everything should be going via the encoders where you limit the possibilities further.

src/riak_dt_multi.erl
lenary: Ah, right, I see, it's in the model. Not pressing, but good to do at some point.

lenary:
Looks good to me!

src/riak_dt_multi.erl
russelldb: I still don't see what functions in `dict` it can use. It uses `dict:update` already. The dictionary is a temp accumulator for all the CRDTs stored in the model under a Key.

lenary: `dict:find(K, Res)` instead of `proplists:get_value(K, dict:to_list(Res))` ?

russelldb: derp, oh yeah!

like I said in hipchat, totally code blind on vvorset and multi

seancribbs:
Some dialyzer warnings:

```
riak_dt_gcounter.erl:80: Invalid type specification for function riak_dt_gcounter:update/3. The success typing is ('increment' | {'increment',pos_integer()},_,riak_dt_gcounter:gcounter()) -> {'ok',riak_dt_gcounter:gcounter()}
riak_dt_multi.erl:289: Function precondition_context/3 will never be called
riak_dt_multi.erl:296: Function precondition_context/2 will never be called
riak_dt_orset.erl:90: Invalid type specification for function riak_dt_orset:update/3. The success typing is ({'add',_} | {'add_all',[any()]} | {'remove',_} | {'remove_all',maybe_improper_list()} | {'update',[any()]},_,_) -> {'error',{'precondition',{_,_}}} | {'ok',_}
riak_dt_orset.erl:90: The return type riak_dt_orset:orset() in the specification of update/3 is not a subtype of {'error',_} | {'ok',_}, which is the expected return type for the callback of riak_dt behaviour
riak_dt_orset.erl:191: The call riak_dt_orset:to_binary({_,[]}) does not have an opaque term of type riak_dt_orset:orset() as 1st argument
riak_dt_orset.erl:196: The return type riak_dt_orset:binary_orset() in the specification of to_binary/1 is not a subtype of binary(), which is the expected return type for the callback of riak_dt behaviour
riak_dt_pncounter.erl:96: Invalid type specification for function riak_dt_pncounter:update/3. The success typing is ('decrement' | 'increment' | {_,integer()},_,_) -> {'ok',_}
riak_dt_vvorset.erl:123: Invalid type specification for function riak_dt_vvorset:update/3. The success typing is ({'add',_} | {'add_all',[any()]} | {'remove',_} | {'remove_all',maybe_improper_list()} | {'update',[any()]},_,_) -> {'error',{'precondition',{_,_}}} | {'ok',_}
riak_dt_vvorset.erl:123: The return type riak_dt_vvorset:vvorset() in the specification of update/3 is not a subtype of {'error',_} | {'ok',_}, which is the expected return type for the callback of riak_dt behaviour
riak_dt_vvorset.erl:267: The call riak_dt_vvorset:to_binary({_,[{_,{_,_}}]}) does not have an opaque term of type riak_dt_vvorset:vvorset() as 1st argument
Unknown types:
  ordsets:ordset/0
  riak_dt_multi:multi_op/0
  riak_dt_orset:orset_op/0
```

seancribbs:
- [x] dialyzer
- [ ] style / organization / typos
- [x] eunit
- [x] eqc
- [ ] documentation

src/riak_dt_multi.erl
```diff
@@ -24,9 +24,32 @@
...
 %% be a CRDT module, yes, even this one.)  The value list is a dict of
-%% {name, type} -> CRDT value mappings.  The value list may have it's
-%% CRDT value removed as soon as the key is removed from the
-%% keyset. The Keyset requries consensus for GC.
+%% {name, type} -> CRDT value mappings.
+%%
+%% The value for a given field is tombstoned when the field is
+%% removed. This map uses a reset / remove strategy to generate
+%% tombstones. Reset attempts to set the value of the CRDT back to
+%% bottom, with out break monotonicity. For example, if a counter at
```
seancribbs: "without breaking"

src/riak_dt_multi.erl
```diff
@@ -24,9 +24,32 @@
...
+%% field F has a value of 10, then on remove of F, the counter is
+%% decremented by 10, so it's value is zero, and the field removed
```
seancribbs: s/it's/its/

src/riak_dt_multi.erl
```diff
@@ -127,56 +156,68 @@ value({contains, Field}, {Schema, _Values}) ->
...
+%% the `multi_op()'. `Ops' is a list of one or more of the following
+%% ops:
+%%
+%% `{update, field(), Op} where `Op' is a valid update operation for a
+%% CRDT of type `Mod' from the `Key' pair `{Name, Mod}' If there is no
+%% local value for `Key' a new CRDT is created, the operation applied
+%% and the result inserted otherwise, the operation is applied to the
+%% local value.
+%%
+%% {add, field}' where field is `{name, type}' results in `field'
```
seancribbs: You used `field()` above, it could be used on this line as well.

src/riak_dt_orset.erl
```diff
@@ -149,9 +172,29 @@ merge_dicts(Dict1, Dict2) ->
...
+%% @Doc reset the set to empty, eseentialy have `Actor' remove all
+%% present members.  Equivalent to update({remove_all,
+%% value(`ORSet')}, `Actor', `ORset')
+-spec reset(orset(), actor()) -> orset().
+reset(ORSet, Actor) ->
+    Values = value(ORSet),
+    {ok, ORSet2} = update({remove_all, Values}, Actor, ORSet),
+    ORSet2.
+
+%% @doc the precondition context is a binary representation of a fragment of the CRDT
```
seancribbs: Not a binary representation really, you're returning an `orset()`.

seancribbs:
@russelldb This doesn't seem to add the partitioning of operations by whether they have preconditions. Is that only in KV or is it here?

seancribbs:
This is really well-tested and clear, even with the many commits. There are only a few documentation typos that need fixing, then :+1:.

russelldb:
@seancribbs yeah, the precondition before / after application is in riak_kv. Reason being that it only makes sense in a hybrid op/state/"action at a distance" context, and not in a pure state based one.

(And I didn't want to complicate the pure CRDTs further.)

PR #54, opened 2013-09-16
Add a tombstoneless OR-Set (ORSWOT) and a Map backed by it

russelldb:
GC looked like too much trouble, so with a bit of thought we came up with a tombstoneless set. Turns out it is just a state based, DVV backed version of this http://arxiv.org/abs/1210.3368. But it was nice to arrive there by our own route and we all understand both DVV and ORSWOT way better now.

I'll gladly squash the commits down after review.


src/riak_dt_gset.erl
```diff
@@ -0,0 +1,141 @@
...
+-export_type([gset/0, binary_gset/0, gset_op/0]).
+-opaque gset() :: members().
+
+-type binary_gset() :: binary(). %% A binary that from_binary/1 will operate on.
+
+-type gset_op() :: {add, member()}.
+
+-type actor() :: riak_dt:actor().
+
+%% an orddict
```
lenary: I don't think this is actually an orddict. Perhaps an ordset?

src/riak_dt_gset.erl
```diff
@@ -0,0 +1,141 @@
...
+
+%% @Doc note: not implemented yet, same as `value/1'
+-spec value(any(), gset()) -> [member()].
+value(_, GSet) ->
+    value(GSet).
+
+
+-spec update(gset_op(), actor(), gset()) -> {ok, gset()}.
+update({add, Elem}, _Actor, GSet) ->
+    {ok, ordsets:add_element(Elem, GSet)}.
```
lenary: A lot of the OR-Set implementations have `{add_all, Elements}`, could this have that too? Or would we never use it. I hate to say "for completeness" as that just makes more for us to maintain.

src/riak_dt_gset.erl
```diff
@@ -0,0 +1,141 @@
...
+
+-spec merge(gset(), gset()) -> gset().
+merge(GSet1, GSet2) ->
+    ordsets:union(GSet1, GSet2).
+
+-spec equal(gset(), gset()) -> boolean().
+equal(GSet1, GSet2) ->
+    ordsets:is_subset(GSet1, GSet2) andalso
+        ordsets:is_subset(GSet2, GSet1).
```
lenary: `GSet1 == GSet2` should work here (because they're ordered, and terms can compare for equality)

src/riak_dt_lwwreg.erl
```diff
@@ -112,11 +112,6 @@ equal({Val, TS}, {Val, TS}) ->
 equal(_, _) ->
     false.
 
-%% @Doc reset to default value
--spec reset(lwwreg(), term()) -> lwwreg().
-reset({_Value, _TS}, _Actor) ->
-    {undefined, 0}.
```
lenary, line 118: For my information, why are we removing `reset` on LWWWWWWWWWWWWs?

russelldb, line 118: No more reset / remove map, so no need for reset on CRDTs.

src/riak_dt_gset.erl
seancribbs: Agree, or if it is an ordset, we should use `ordsets:ordset(Type)` in the type spec.

src/riak_dt_gset.erl
seancribbs: @lenary :+1:

src/riak_dt_gset.erl
seancribbs: @lenary :+1:

src/riak_dt_map.erl
```diff
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% riak_dt_multi: OR-Set schema based multi CRDT container
+%% riak_dt_map: OR-Set schema based multi CRDT container
```
seancribbs, line 3: :clap: :clap: :clap:

test/crdt_statem_eqc.erl
```diff
@@ -133,7 +133,8 @@ crdt_equals(Mod, {_IDS, CS}, {_IDD, CD}) ->
 %% Helpers
 %% The orset CRDT returns a list, it has no guarantees about order
 %% list equality expects lists in order
-sort(Mod, L) when Mod == riak_dt_vvorset; Mod == riak_dt_multi  ->
+sort(Mod, L) when Mod == riak_dt_vvorset; Mod == riak_dt_multi; Mod == riak_dt_orset;
```
seancribbs: You should remove `riak_dt_vvorset` here if the module is deleted.

seancribbs:
Re: `riak_dt_map`, I realize it is likely more efficient in the serialized form, but do we really need both the ORSWOT of keys and the orddict of key-value pairs? Do we need to reuse the ORSWOT or can we just make it use similar merge logic? It feels to me that entries of `{Key, Dot, Value}` would be more space-efficient and easier to reason about.

seancribbs:
- [x] eunit
- [x] quickcheck
- [ ] dialyzer
- [ ] style / substance / nitpicks

seancribbs:
Dialyzer found these:

```
riak_dt_orswot.erl:198: The call riak_dt_vclock:replace_actors([{_,_}],[{_,_},...]) does not have an opaque term of type riak_dt_vclock:vclock() as 2nd argument
riak_dt_orswot.erl:222: The call riak_dt_orswot:vclock_merge([any(),...],[any(),...],Actors1::[{_,integer()}],Actors2::[{_,integer()}],MergedActors::[{_,integer()}]) does not have opaque terms as 1st and 2nd arguments
riak_dt_orswot.erl:228: Function greatest_dot/1 will never be called
Unknown types:
  riak_dt_orswot:orswot/1
```
- The errors on lines 198 and 222 come from treating the `vclock` type as non-opaque; i.e. composing it from other types without using the module. Perhaps this would be a case where `riak_dt_vclock` can export some appropriate functions that operate on its types?
- The unknown type originates from riak_dt_map.erl line 59, the `orswot` type is exported as 0-arity.

src/riak_dt_orswot.erl
```diff
@@ -0,0 +1,423 @@
...
+remove_all([Elem | Rest], Actor, ORSet) ->
+    case update({remove, Elem}, Actor, ORSet) of
+        {ok, ORSet2} ->
+            remove_all(Rest, Actor, ORSet2);
+        Error ->
+            Error
+    end.
+
+-spec merge(orswot(), orswot()) -> orswot().
+merge({Clock1, LHSActors, LHSEntries}, {Clock2, RHSActors, RHSEntries}) ->
```
seancribbs: It seems to me the logic of `merge/2` is getting obscured and complicated by the actor-replacement/compression stuff. Why not compute the actor-replacements when serializing/deserializing instead? Then you could optimize cases where, say, the clock of one ORSWOT strictly dominates the other. You would also simplify the rest of the merge logic by using the actual actor IDs.

src/riak_dt_orswot.erl
```diff
@@ -0,0 +1,423 @@
...
+                              [NewDot] = riak_dt_vclock:replace_actors(orddict:to_list(MergedActors), [{Real, Count}]),
+                              orddict:store(Key, NewDot, Acc);
+                          true ->
+                              Acc
+                      end
+              end,
+              Accumulator,
+              Keys).
+
+has_seen_dot({Actor, ClockCount}, {Actor, DotCount}) when ClockCount >= DotCount ->
```
seancribbs: Can this be part of `riak_dt_vclock`? Maybe a variant of `descends`?

src/riak_dt_orswot.erl
```diff
@@ -0,0 +1,423 @@
...
+                      V2 = orddict:fetch(Key, Entries2),
+                      V = vclock_merge([V1], [V2], Actors1, Actors2, MergedActors),
+                      NewDot = greatest_dot(V),
+                      orddict:store(Key, NewDot, Acc) end,
+              orddict:new(),
+              CommonKeys).
+
+greatest_dot([Dot]) ->
+    Dot;
+greatest_dot([{A1, C1}, {_A2, C2}]) when C1 > C2 ->
```
seancribbs: I'm concerned about this logic. Basically it says, even if an actor with a smaller count wrote after another, it'll choose the other/subsume that. Am I reading that correctly?

src/riak_dt_orswot.erl
```diff
@@ -0,0 +1,423 @@
...
+-ifdef(TEST).
+
+-ifdef(EQC).
+eqc_value_test_() ->
+    crdt_statem_eqc:run(?MODULE, 1000).
+
+generate() ->
+    ?LET(Members, list(int()),
+         lists:foldl(fun(M, Set) ->
+                            riak_dt_orswot:update({add, M}, choose(1, 50), Set) end,
```
seancribbs, line 267: No big deal, but if we want to enforce binary actor IDs later, this will break from the `choose(1,50)`.

src/riak_dt_orswot.erl
```diff
@@ -0,0 +1,423 @@
...
+%% EQC generator
+gen_op() ->
+    oneof([gen_updates(), gen_update()]).
+
+gen_updates() ->
+     {update, non_empty(list(gen_update()))}.
+
+gen_update() ->
+    oneof([{add, int()}, {remove, int()},
+           {add_all, list(int())},
```
seancribbs: Use `non_empty()` on these generators.

src/riak_dt_orswot.erl
```diff
@@ -0,0 +1,423 @@
...
+           {remove_all, list(int())}]).
+
+init_state() ->
+    {0, dict:new()}.
+
+do_updates(_ID, [], _OldState, NewState) ->
+    NewState;
+do_updates(ID, [{_Action, []} | Rest], OldState, NewState) ->
```
seancribbs: When you have `non_empty()` on lines 333/334, you can remove this clause.

src/riak_dt_orswot.erl
russelldb: Can you explain your concern?

src/riak_dt_orswot.erl
```diff
@@ -0,0 +1,423 @@
...
+    do_updates(ID, Rest, OldState, NewState);
+do_updates(ID, [Update | Rest], OldState, NewState) ->
+    case {Update, update_expected(ID, Update, NewState)} of
+        {{Op, _Arg}, NewState} when Op == remove;
```
seancribbs, line 290: Can you explain this clause to me?

russelldb, line 290: If the op was a remove and the model state didn't change, then we're looking at a precondition error, so no operations would have been applied to the orset state.

src/riak_dt_orswot.erl
russelldb: I hadn't thought of that, but it makes sense. Good idea, thanks.

src/riak_dt_orswot.erl
russelldb: Makes sense, yes, I'll try that.

src/riak_dt_orswot.erl
russelldb: I'm having a think about this. I can see how it is OK for a pair of replicas, I need to think it through for a larger number. Maybe you're right to be concerned. The alternative of course is to have both dots…I'll get back to you. I think I can see the problem.

A, B and C. A and B concurrently add 'bob', so in A bob->{a,1} and in B bob->{b, 1}.
B merges with C so in C bob->{b, 1}
B merges with A and we pick B's dot bob->{b, 1}
C removes bob
C merges with A and since C's clock must have an entry for {b, 1} 'bob' is out of the set. Yet if the entry was {a, 1} bob would be in.

This suggests we need more than the dot.

Yeah, Hmmm. 'Cos if A merged with C, then B with A, then C removed bob and merged with A, bob would be in. I wonder if my quickcheck tests are any use at all.

seancribbs: Ah yes, that's a very simple example where it fails. Can we confirm with @CBaquero?

src/riak_dt_orswot.erl
seancribbs, line 290: Thanks, that makes more sense now.

src/riak_dt_orswot.erl
russelldb: We spoke about it on Friday and he was pretty convincing. I've emailed him. It is as though C never gets the message about A's add. But does it matter?

seancribbs:
I think I'm done with this pass of review. Those dialyzer fixes are probably minor and don't impede on the rest of the code. I'll take another pass after you've been able to address the comments.

russelldb:
Cool, thanks Sean, some stuff to think about there.

src/riak_dt_orswot.erl
seancribbs: If this logic ends up being ok, the other option would be to generalize this into a `max_dot/2` function and just fold over the list of dots.

russelldb: Yeah, I'm going over it in my head, and re-reading the orswot spec. I think the confusion could be that the spec stores (element, count, actor) where as we _were_ storing (element, [{replica, count}]). It doesn't have to be the whole version vector, but it does need to grow when we merge to encompass concurrent adds. You spotted a good one here. Thanks.

We really need to get some better tests as well, to ensure we catch problems like this.

test/crdt_statem_eqc.erl
russelldb: Oh ya And Multi, too.

russelldb:
I've addressed all the comments except the actor list in orswot and map using orswot.
1. I'd like to address the actor list comment by doing the to/from binary work
2. I'd like you to look at the map here https://github.com/basho/riak_dt/compare/feature/rdb-more-orset...feature/rdb-map2?expand=1 and see what you think of that. It uses similar logic to orswot, but has a more fine grained approach to causality by keeping sibling fields on merge.

russelldb:
OK, Pretty sure I've addressed all your comments. I'll start the to/from binary stuff in a new branch for pre3 when this is merged. (I'll open issues for any unaddressed comments as per Jon's instructions.) cc @seancribbs / @lenary (nag nag hint hint pre2 is Friday)

seancribbs:
@russelldb Sounds good, doing a final pass now.

seancribbs:
Minor typo I found in riak_dt_map. I'm going to make that change since it's late UK time, then you can merge at will. :+1:

russelldb:
Thanks you @seancribbs!

russelldb:
Squashed and merged.

Issue #55, opened 2013-09-25
Other Custom Data Types [JIRA: RIAK-2714]


lenary:
Sean has suggested an lmax (useful for gaming), and I think we had a few others in the works (lists etc).
This is a placeholder for the moment, but we can look at adding these.
- [ ] Max Int / Min Int
- [ ] Sorted/Scored set (Z-set)

seancribbs:
Assigned 2.1 milestone

tcrossland:
I would also suggest the following:
- [ ] PN-Counter

russelldb:
@tcrossland another one than this one https://github.com/basho/riak_dt/blob/develop/src/riak_dt_pncounter.erl ?

tcrossland:
Ah, sorry, I missed it in the docs but indeed it exists. Thanks for pointing it out.

russelldb:
np

PR #59, opened 2013-09-30
A simple MaxReg implementation

lenary:
In accordance with prophecy #55, I have written a register that stores the largest integer that you assign to it.

Questions
- [x] Should this be for all terms, or just numbers? (I think it's easier to manage expectations with just numbers)
- [x] Should it be integers or all numbers? (I coded with integers to start now, easily expanded)
- [ ] What do we do about returning `undefined` to a user? 
- [x] Do we error if someone assigns a smaller number to the register than what's in it? (I think no)
- [x] Do we error if someone assigns a non-number to the register? (I think yes)
- [x] Do we want a MinReg too? (I think maybe, but one was enough to start)

### DIFF
--- a/src/riak_dt_maxreg.erl
+++ b/src/riak_dt_maxreg.erl
@@ -28,9 +28,11 @@
 %% @end
 
 -module(riak_dt_maxreg).
+-behaviour(riak_dt).
 
--export([new/0, value/1, value/2, update/3, merge/2,
-         equal/2, to_binary/1, from_binary/1]).
+-export([new/0, value/1, value/2, update/3, update/4, merge/2,
+         equal/2, to_binary/1, from_binary/1, parent_clock/2,
+         stats/1, stat/2]).
 
 %% EQC API
 -ifdef(EQC).
@@ -48,6 +50,8 @@
 
 -type maxreg_op() :: {assign, integer()}.
 
+-type nonintegral_error() :: {error, {type, {nonintegral, term()}}}.
+
 %% @doc Create a new, empty `maxreg()'
 -spec new() -> maxreg().
 new() ->
@@ -64,10 +68,22 @@ value(_, V) ->
     value(V).
 
 %% @doc Assign a `Value' to the `maxreg()'
--spec update(maxreg_op(), term(), maxreg()) ->
-                    {ok, maxreg()}.
+-spec update(maxreg_op(), riak_dt:actor(), maxreg()) ->
+                    {ok, maxreg()} | nonintegral_error().
 update({assign, Value}, _Actor, OldVal) when is_integer(Value) ->
-    {ok, merge(OldVal, Value)}.
+    {ok, merge(OldVal, Value)};
+update({assign, Value}, _Actor, _OldVal) ->
+    {error, {type, {nonintegral, Value}}}.
+
+-spec update(maxreg_op(), riak_dt:actor(), maxreg(), riak_dt:context()) ->
+                    {ok, maxreg()} | nonintegral_error().
+update(Op, Actor, MaxReg, _Ctx) ->
+    update(Op, Actor, MaxReg).
+
+-spec parent_clock(riak_dt_vclock:vclock(), maxreg()) ->
+                          maxreg().
+parent_clock(_Clock, MaxReg) ->
+    MaxReg.
 
 %% @doc Merge two `maxreg()'s to a single `maxreg()'. This is the Least Upper Bound
 %% function described in the literature.
@@ -102,6 +118,15 @@ to_binary(MaxReg) ->
 -spec from_binary(binary()) -> maxreg().
 from_binary(<<?TAG:8/integer, ?V1_VERS:8/integer, Bin/binary>>) ->
     riak_dt:from_binary(Bin).
+
+
+%% @doc No Stats because it's just an integer
+-spec stats(maxreg()) -> [{atom(), number()}].
+stats(_MaxReg) -> [].
+
+%% @doc No Stats because it's just an integer
+-spec stat(atom(), maxreg()) -> number() | 'undefined'.
+stat(_, _) -> undefined.
 
 %% ===================================================================
 %% EUnit tests
--- a/src/riak_dt_maxreg.erl
+++ b/src/riak_dt_maxreg.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_dt_maxreg: A Register that stores the largest integer assigned into it
 %%
-%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2014 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -83,9 +83,8 @@ max_with_small_undefined(X, undefined) ->
 max_with_small_undefined(X, Y) ->
     max(X, Y).
 
-%% @doc Are two `maxreg()'s structurally equal? This is not `value/1' equality.
-%% Two registers might represent the value `armchair', and not be `equal/2'. Equality here is
-%% that both registers contain the same value and timestamp.
+%% @doc Are two `maxreg()'s structurally equal? Equality here is
+%% that both registers contain the same value.
 -spec equal(maxreg(), maxreg()) -> boolean().
 equal(Val1, Val2) ->
     Val1 =:= Val2.
@@ -94,6 +93,7 @@ equal(Val1, Val2) ->
 -define(V1_VERS, 1).
 
 % TODO: is 32 bits of integer enough? Do we require more bits?
+% TODO: do we want to be using t2b/b2t here?
 
 %% @doc Encode an effecient binary representation of an `maxreg()'
 -spec to_binary(maxreg()) -> binary().
--- a/src/riak_dt_maxreg.erl
+++ b/src/riak_dt_maxreg.erl
@@ -50,8 +50,6 @@
 
 -type maxreg_op() :: {assign, integer()}.
 
--type nonintegral_error() :: {error, {type, {nonintegral, term()}}}.
-
 %% @doc Create a new, empty `maxreg()'
 -spec new() -> maxreg().
 new() ->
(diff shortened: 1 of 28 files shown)


src/riak_dt_maxreg.erl
```diff
@@ -0,0 +1,143 @@
...
+max_with_small_undefined(undefined, X) ->
+    X;
+max_with_small_undefined(X, undefined) ->
+    X;
+max_with_small_undefined(X, Y) ->
+    max(X, Y).
+
+%% @doc Are two `maxreg()'s structurally equal? This is not `value/1' equality.
+%% Two registers might represent the value `armchair', and not be `equal/2'. Equality here is
+%% that both registers contain the same value and timestamp.
```
russelldb: Looks like a copypasta doc comment from LWW?

russelldb:
In answer to you questions
1. numbers
2. integers
3. I dunno, what's wrong with zero (misleading in a usecase with negative numbers I suppose)
4. No
5. Yes
6. Yes please!

seancribbs:
Maybe a more general "Range register"? Something that records the min, max, and latest value?

lenary:
@rdb for 3, I wonder if we can return some kind of negative infinity. I have no idea, going to stick with undefined for the moment. As for other answers, I agree, will sort out everything. :D 

@seancribbs I'll start working on that in a second. Useful for finance stuff, where you want (max,min,earliest,latest)

lenary:
Needs to be rebased against #71 due to tag changes.

seancribbs:
Assigning to 2.1

lenary:
As expected.

src/riak_dt_maxreg.erl
```diff
@@ -0,0 +1,144 @@
...
+
+-ifdef(EQC).
+eqc_value_test_() ->
+    crdt_statem_eqc:run(?MODULE, 1000).
+
+%% EQC generator
+generate() ->
+    ?LET({Op, Actor}, {gen_op(), char()},
+         begin
+             {ok, Lww} = riak_dt_maxreg:update(Op, Actor, riak_dt_maxreg:new()),
```
russelldb: lol, copy pasta naming of the register as `Lww`

src/riak_dt_maxreg.erl
```diff
@@ -0,0 +1,144 @@
...
+-export([gen_op/0, update_expected/3, eqc_state_value/1, init_state/0, generate/0]).
+-endif.
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-export_type([maxreg/0, maxreg_op/0]).
+
+-opaque maxreg() :: integer() | undefined.
```
russelldb: We can't do this, `opaque` are treated as unique by dialyzer. This claims the type `integer() | undefined` uniquely for `maxreg()`.

src/riak_dt_maxreg.erl
```diff
@@ -0,0 +1,144 @@
...
+-spec equal(maxreg(), maxreg()) -> boolean().
+equal(Val1, Val2) ->
+    Val1 =:= Val2.
+
+-include("riak_dt_tags.hrl").
+-define(TAG, ?DT_MAXREG_TAG).
+-define(V1_VERS, 1).
+
+% TODO: is 32 bits of integer enough? Do we require more bits?
+% TODO: do we want to be using t2b/b2t here?
```
russelldb: I don't think we need to.

src/riak_dt_maxreg.erl
```diff
@@ -0,0 +1,144 @@
...
+%% that both registers contain the same value.
+-spec equal(maxreg(), maxreg()) -> boolean().
+equal(Val1, Val2) ->
+    Val1 =:= Val2.
+
+-include("riak_dt_tags.hrl").
+-define(TAG, ?DT_MAXREG_TAG).
+-define(V1_VERS, 1).
+
+% TODO: is 32 bits of integer enough? Do we require more bits?
```
russelldb: Hmmm. I don't know, if I say yes, no doubt there will be a slap down the line, so maybe use t2b and it won't matter!

lenary: now that riak_dt has helpers for this, it's pretty simple to use t2b and b2t. I'll get that done.

src/riak_dt_maxreg.erl
russelldb: Actually, I'm unsure of this now, I need to check with @reiddraper, who first brought this to my attention. Hold off for now.

src/riak_dt_minreg.erl
```diff
@@ -0,0 +1,144 @@
...
+%% that both registers contain the same value.
+-spec equal(minreg(), minreg()) -> boolean().
+equal(Val1, Val2) ->
+    Val1 =:= Val2.
+
+-include("riak_dt_tags.hrl").
+-define(TAG, ?DT_MINREG_TAG).
+-define(V1_VERS, 1).
+
+% TODO: is 32 bits of integer enough? Do we require more bits?
```
russelldb: same as for `maxreg`, I guess you know that now.

src/riak_dt_minreg.erl
```diff
@@ -0,0 +1,144 @@
...
+
+-ifdef(EQC).
+eqc_value_test_() ->
+    crdt_statem_eqc:run(?MODULE, 1000).
+
+%% EQC generator
+generate() ->
+    ?LET({Op, Actor}, {gen_op(), char()},
+         begin
+             {ok, Lww} = riak_dt_minreg:update(Op, Actor, riak_dt_minreg:new()),
```
russelldb: as for `maxreg` I know it is a small thing, but it is confusing on first read.

src/riak_dt_rangereg.erl
```diff
@@ -0,0 +1,241 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc
```
russelldb: All copy pasta, please add a real `@doc` comment

src/riak_dt_rangereg.erl
```diff
@@ -0,0 +1,241 @@
...
+fww(undefined, Other) ->
+  Other;
+fww(Other, undefined) ->
+  Other;
+fww({VA,TsA}, {_VB,TsB}) when TsA < TsB ->
+  {VA,TsA};
+fww({_VA,TsA}, {VB,TsB}) when TsA > TsB ->
+  {VB,TsB};
+fww(A, B) ->
+  min(A,B). %% some kind of tiebreaker.
```
russelldb: Please document this arbitrator in either `merge/2` or the module edoc.

src/riak_dt_rangereg.erl
```diff
@@ -0,0 +1,241 @@
...
+%% that both registers contain the same value.
+-spec equal(rangereg(), rangereg()) -> boolean().
+equal(Val1, Val2) ->
+    Val1 =:= Val2.
+
+-include("riak_dt_tags.hrl").
+-define(TAG, ?DT_RANGEREG_TAG).
+-define(V1_VERS, 1).
+
+% TODO: compression?
```
russelldb: Please measure briefly, I doubt there is any value to compressing vs. cpu time, but why not check?

src/riak_dt_rangereg.erl
```diff
@@ -0,0 +1,241 @@
...
+
+-ifdef(EQC).
+eqc_value_test_() ->
+    crdt_statem_eqc:run(?MODULE, 1000).
+
+%% EQC generator
+generate() ->
+    ?LET({Op, Actor}, {gen_op(), char()},
+         begin
+             {ok, Lww} = riak_dt_rangereg:update(Op, Actor, riak_dt_rangereg:new()),
```
russelldb: What's in a name?

src/riak_dt_rangereg.erl
lenary: Done

src/riak_dt_rangereg.erl
lenary: Done both

lenary:
TODO:
- [x] stats callbacks (required by kv)
- [x] error if non-integral value is assigned to the register.

lenary:
Ok, I think the last things to do are to benchmark the compression (vs without), and to look at some interface details.

lenary:
Benchmarking compression (via `riak_dt:to_binary`) vs nowt (via `term_to_binary`):

```
         compd        uncompd
maxreg (and minreg):
time:    4.04         0.39
size:    [5..8]       [5..8]

rangereg:
time:    9.49         0.90
size:    [32..43]     [32..44]
```

so i guess no compression then

src/riak_dt_maxreg.erl
seancribbs: @russelldb You are correct. Opaque types must be composite types (records usually). They can't be a union type of two or more fundamental types.

src/riak_dt_maxreg.erl
```diff
@@ -0,0 +1,162 @@
...
+-spec value(term(), maxreg()) -> term().
+value(_, V) ->
+    value(V).
+
+%% @doc Assign a `Value' to the `maxreg()'
+-spec update(maxreg_op(), riak_dt:actor(), maxreg()) ->
+                    {ok, maxreg()} | nonintegral_error().
+update({assign, Value}, _Actor, OldVal) when is_integer(Value) ->
+    {ok, merge(OldVal, Value)};
+update({assign, Value}, _Actor, _OldVal) ->
```
russelldb: I wonder about this, what do you think @seancribbs? For me, erlang would generally not have this case, and just let the caller crash.

russelldb:
Most CRDTs also throw in the to/from binary roundtrip test (see https://github.com/basho/riak_dt/blob/develop/src/riak_dt_orswot.erl#L562), would you mind, awfully?

src/riak_dt_maxreg.erl
seancribbs: You could let the caller crash, yes, or use `error(badarg)`.

russelldb:
I'm plus one on this, but I'm not going to plus one until we branch for 2.0rc.

Good job, man! Welcome back!

PR #62, opened 2013-10-01
To / from binary for map and set

russelldb:
The eqc tests collect output of the format

`{{size range}, {percent saving rang}}`

Where size is `{from, to}` eg from set it is the cardinality, so `{10, 20}` means a set with 10-20 elements.
For map size is the `term_to_binary` size `div 10`.

Percent saving of `{20, 30}` means that the `to_binary/1` output is 20-30% smaller than t2b for the same data.

Only when the data type is empty is t2b better (as far as I can tell.)

### DIFF
--- a/src/riak_dt_map.erl
+++ b/src/riak_dt_map.erl
@@ -49,7 +49,7 @@
 %% EQC API
 -ifdef(EQC).
 -export([gen_op/0, update_expected/3, eqc_state_value/1,
-         init_state/0, gen_field/0]).
+         init_state/0, gen_field/0, generate/0, size/1]).
 -endif.
 
 -export_type([map/0, binary_map/0, map_op/0]).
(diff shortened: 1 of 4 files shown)


src/riak_dt_vclock.erl
```diff
@@ -147,8 +148,20 @@ all_nodes(VClock) ->
...
     lists:sort(VA) =:= lists:sort(VB).
 
+actors(Clock) ->
+    [Actor || {Actor, _Count} <- sort(Clock)].
+
+sort(Clock) ->
+    lists:sort(Clock).
+
+to_binary(Clock) ->
+    term_to_binary(sort(Clock)).
```
lenary, line 161: Can we do better here? Is there much point? Can we at least give the thing a tag and a version number?

src/riak_dt_map.erl
```diff
@@ -273,28 +273,167 @@ precondition_context(Map) ->
...
 -spec from_binary(binary_map()) -> map().
-from_binary(<<?TAG:8/integer, ?V1_VERS:8/integer, Bin/binary>>) ->
-    %% @TODO something smarter
+from_binary(<<?TAG:8/integer, ?V1_VERS:8/integer,
+              ModMapBinLen:32/integer, ModMapBin:ModMapBinLen/binary,
+              MapBin/binary>>) ->
+    ModMap = mod_map_from_binary(ModMapBin),
+    from_binary(MapBin, ModMap).
+
+to_binary({Clock, Fields}, ModMap0) ->
```
lenary, line 331: If we go with the version of to_binary that supports upgrades (`to_binary(CRDT, WantedVersion)`), this function, and the current `from_binary/2` will need to change.

seancribbs:
Review TODOs: 
- [x] eunit / eqc passes
- [ ] dialyzer passes
- [x] style / formatting / nitpicks
- [ ] documentation

src/riak_dt_vclock.erl
russelldb, line 161: Not from my measurement. I reckon implicit versioning `<<131…>>` means v1 :D

src/riak_dt_map.erl
russelldb, line 331: Hmmm. I agree. And they can change then.

seancribbs:
Some dialyzer warnings:

```
riak_dt_map.erl:322: The call riak_dt_vclock:to_binary(Clock::riak_dt_vclock:vclock()) contains an opaque term as 1st argument when a structured term of type [any()] is expected
riak_dt_orswot.erl:264: Invalid type specification for function riak_dt_orswot:to_binary/1. The success typing is ({[any()],[{_,riak_dt_vclock:vclock()}]}) -> <<_:8,_:_*1>>
riak_dt_orswot.erl:285: The call riak_dt_vclock:to_binary(Clock::riak_dt_vclock:vclock()) contains an opaque term as 1st argument when a structured term of type [any()] is expected
riak_dt_orswot.erl:295: Invalid type specification for function riak_dt_orswot:from_binary/1. The success typing is (<<_:8,_:_*1>>) -> {[any()],[{_,_}]}
```

I think all of these can be resolved by adding specs for `riak_dt_vclock:to_binary/1` and `riak_dt_vclock:from_binary/1`. The main issue is that because `vclock()` is an opaque type, dialyzer can't determine that `[any()]` (the inferred type) actually matches it.

src/riak_dt_orswot.erl
```diff
@@ -243,42 +262,106 @@ precondition_context(ORSet) ->
...
+from_binary(<<?TAG:8/integer, ?V1_VERS:8/integer, ClockLenFieldLen:8/integer,
+            ClockLen:ClockLenFieldLen/integer, BinClock:ClockLen/binary, BinEntries/binary>>) ->
+    Clock = riak_dt_vclock:from_binary(BinClock),
+    Actors0 = riak_dt_vclock:actors(Clock),
+    Actors = lists:zip(lists:seq(1, length(Actors0)), Actors0),
+    Entries = binary_to_entries(BinEntries, ClockLenFieldLen, Actors, orddict:new()),
+    {Clock, Entries}.
+
+binary_to_entries(<<>>, _, _, Entries) ->
+ Entries;
```
seancribbs, line 340: Some minor whitespace niggle here, indentation looks weird.

seancribbs:
In general this is great, nice space savings for larger data structures.

I think we can do better on documentation, but maybe that deserves an issue to itself? Also it would be nice to see that roundtrip applied to all types.

Ping again when you have those dialyzer warnings squelched and I'll give it another pass and the thumbs-up.

russelldb:
Thanks @seancribbs. I'll get it cleaned up UK morning.

lenary:
@seancribbs I want to get my test back in the running at some stage, which also does a full roundtrip. Let's just see how we go.

seancribbs:
Russell's prop does round trips and measures the space savings. We just need to implement the callbacks in each module. pre4 stuff IMHO.

Sean Cribbs

lenary:
@seancribbs yeah, that comment wasn't "block this PR until i get mine back in the running", it was "i should do this sometime, in a separate PR"

russelldb:
ping @seancribbs addressed comments and made dialyzer happy. Also noticed that the `actors/1` fun in `riak_dt_vclcok` was duplicate of the already present `all_nodes/1` fun, so removed it.

seancribbs:
Still one lingering dialyzer warning:

```
Unknown types:
  riak_dt_vclock:actor/0
```

Here's where that occurs:

```
src/riak_dt_map.erl
97:-type actor_map() :: [{riak_dt_vclock:actor(), pos_integer()}
98:                      | {pos_integer(), riak_dt_vclock:actor()}].
```

seancribbs:
Once that type warning is fixed, :+1: to merge.

russelldb:
Thanks, I made issue #63 to track adding the roundtrip test to other types.

PR #64, opened 2013-10-11
Improve efficiency of flags.

seancribbs:
This changes the OE/OD flags to be "tombstone-less", like ORSWOTs (or more accurately, VVORSets). Basically the "observed" operation flips the flag without incrementing, while the opposite operation increments the vector clock. This greatly reduces space complexity, and consolidates our garbage into the same type as produced by other types, namely actor IDs.

I also put the binary-roundtrip EQC into both of these types and added type annotations.

@russelldb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ dev/*
 doc/*
 rel/riak_dt
 data/*
+.qc/*
+.eqc-info
+current_counterexample.eqc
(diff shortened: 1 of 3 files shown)


russelldb:
Tests pass, code reads very well. Sneaking suspicion these flags could be a single module, but maybe the added complexity is worse than the slight duplication. +1, merge at will. And thanks!

PR #67, opened 2013-11-11
Added assign w/o timestamp top the lwwreg_op.

Licenser:
{assign, Value} is a valid lwwreg_op at this time, it should be declared as such to make dialyzer happy.


russelldb:
:+1: many thanks!

PR #69, opened 2013-11-20
Add a stats/1 callback to all datatypes.

seancribbs:
The callback returns a proplist of atom/number pairs about internal statistics of the datatype. Examples include 'actor_count' (on any VV-based type), 'element_count' (sets), 'field_count' (maps), 'max_dot_length' (on dotted VV-based types).

Still undecided: whether to recurse into embedded values for internal stats.

### DIFF
--- a/src/riak_dt.erl
+++ b/src/riak_dt.erl
@@ -38,6 +38,7 @@
 -callback equal(crdt(), crdt()) -> boolean().
 -callback to_binary(crdt()) -> binary().
 -callback from_binary(binary()) -> crdt().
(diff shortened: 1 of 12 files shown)


lenary:
:+1: On all of this.

Maybe if they call `stats([recursion])` then we show recursive stats for the map, but that's going to be fairly expensive, so should be off by default. Getting the riak_object size should be cheap and give the user essentially as much information.

russelldb:
Looks good. Does enough for 2.0 purposes. :+1:

seancribbs:
Going to add some more -spec annotations and then merge.

PR #73, opened 2013-12-03
Add stat/2 callback for retrieving individual stats.

seancribbs:
This will allow us to configure specific stats to be tracked in KV, and only extract those stats we want.

### DIFF
--- a/src/riak_dt_orset.erl
+++ b/src/riak_dt_orset.erl
@@ -117,32 +117,40 @@ precondition_context(ORDict) ->
 
 -spec stats(orset()) -> [{atom(), number()}].
 stats(ORSet) ->
-    {Tags, Tombs} = orddict:fold(fun(_K, {A, R}, {As, Rs}) ->
-                                         {length(A) + As, length(R) + Rs}
-                                 end, {0,0}, ORSet),
-    [
-     {element_count, orddict:size(ORSet)},
-     {adds_count, Tags},
-     {removes_count, Tombs},
-     {waste_pct, Tombs / Tags * 100}
-    ].
+    [ {S, stat(S, ORSet)} || S <- [element_count,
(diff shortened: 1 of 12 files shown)


src/riak_dt_orset.erl
```diff
@@ -127,6 +127,24 @@ stats(ORSet) ->
...
+stat(removes_count, ORSet) ->
+    orddict:fold(fun(_K, {_, R}, Acc) ->
+                         Acc + length(R)
+                 end, 0, ORSet);
+stat(waste_pct, ORSet) ->
+    {Tags, Tombs} = orddict:fold(fun(_K, {A, R}, {As, Rs}) ->
+                                         {length(A) + As, length(R) + Rs}
+                                 end, {0,0}, ORSet),
+    Tombs / Tags * 100;
+stat(_, _) -> undefined.
```
lenary, line 154: You've not changed `stats/1` in this module, so there's duplication of code. Was that intentional?

seancribbs, line 154: I debated about that, since three stats are computed from the same operation. We're not using this module in production, so maybe it doesn't matter as much?

lenary, line 154: I just wanted to make sure it wasn't an oversight.

lenary:
:+1: I'm happy with this.

seancribbs:
@russelldb Awaiting your approval, good sir.

russelldb:
No tests, but the changes look sane. :+1: but I'd like an issue to add eunit tests after the code freeze.

seancribbs:
@russelldb I added the eunits. @lenary I addressed the code duplication and some other bugs in `riak_dt_orset`.

russelldb:
WFM now :+1:

PR #76, opened 2014-02-03
Fix type specification.

cmeiklejohn:
This type shouldn't be the general Erlang term, it should be the opaque
term type.


russelldb:
+1.

PR #77, opened 2014-02-11
Our to/from binary was slow slow slow, use t2b (with compression)

russelldb:
See results here for facile test https://gist.github.com/russelldb/d330d796ca1b25d1879d

The `compression` option is configurable. Defaults to `1`. I still need to cuttlefish this, or at least add it to advanced.config for riak.

### DIFF
--- a/src/riak_dt.app.src
+++ b/src/riak_dt.app.src
@@ -11,7 +11,7 @@
   %% @See docs for erlang:term_to_binary/2. Compression can be either
   %% true | false | integer() when integer() is in a range of 0 - 9. 0
   %% is the same as false, 6 is the same as true.  The integer
-  %% indeicates the level of compression. Higher number means more
+  %% indicates the level of compression. Higher number means more
   %% compression, but more time to compress. In tests so far 1 has
   %% been enough for CRDTs
   {env, [{binary_compression, 1}]}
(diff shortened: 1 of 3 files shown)


src/riak_dt_map.erl
```diff
@@ -311,180 +304,31 @@ stat(_,_) -> undefined.
...
-    <<?TAG:8/integer, ?V1_VERS:8/integer,
-      ModMapBinLen:32/integer,
-      ModMapBin:ModMapBinLen/binary,
-      MapBin/binary>>.
+    Opts = case application:get_env(riak_dt, binary_compression, true) of
+               true -> [compressed];
+               N when N >= 0, N =< 9 -> [{compressed, N}];
+               _ -> []
+           end,
+    <<?TAG:8/integer, ?V1_VERS:8/integer, (term_to_binary(Map, Opts))/binary>>.
```
seancribbs, line 322: I suppose we're not guaranteeing backwards compatibility in formats until a tag?

src/riak_dt.app.src
```diff
@@ -8,5 +8,5 @@
                   kernel,
                   stdlib
                  ]},
-  {env, []}
+  {env, [{binary_compression, 1}]}
```
seancribbs, line 17: I know what this means, but a little documentation here might help. Do we also want to have a cuttlefish schema?

seancribbs:
:+1: This is a must-have in my opinion. We should open an additional PR that adds a cuttlefish schema, happy to do that for you.

PR #71, opened 2014-02-18
Move all version / tag information to a single location

lenary:
Currently the to/from binary stuff is per file, the macros for version / tag should be in a single location, if just because adding a new one means looking in every file.

riak_dt.hrl seems like the logical place.

maybe a macro like ?TAG(?MODULE) would do it?

### DIFF
--- a/include/riak_dt_tags.hrl
+++ b/include/riak_dt_tags.hrl
@@ -1,0 +1,35 @@
+%%% Tags for all our riak_dt CRDTs.
+%%%
+%%% Versions still live in the files themselves, allowing each data type to
+%%% decide when it's a different version on its own.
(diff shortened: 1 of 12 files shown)


lenary:
Mine!

russelldb:
Utterly fantastic! Thanks. (+1 from me)

lenary:
Ok, I added a little more docs. Going to mandate that new tags are checked out of the 90+ range, and should be at least tried to be kept organised...

seancribbs:
Yeah, let's merge this.

russelldb:
make test/dialyzer/xref all pass. Many thanks @lenary, much need, good work!

PR #79, opened 2014-02-18
The attempt to optimise merge was incorrect

russelldb:
If LHS vclock dominates RHS vclock, there can still be removes
that need to be observed, so that optimisation was bogus.

### DIFF
--- a/src/riak_dt_orswot.erl
+++ b/src/riak_dt_orswot.erl
@@ -172,41 +172,25 @@ remove_all([Elem | Rest], Actor, ORSet) ->
 -spec merge(orswot(), orswot()) -> orswot().
 merge({Clock, Entries}, {Clock, Entries}) ->
     {Clock, Entries};
-merge({LHSClock, LHSEntries}=LHS, {RHSClock, RHSEntries}=RHS) ->
-    case either_dominates(LHSClock, RHSClock) of
-        LHSClock -> LHS ;
-        RHSClock -> RHS;
-        concurrent ->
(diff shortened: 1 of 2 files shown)


russelldb:
Hat tip to https://github.com/asonge for the find, and the test.

src/riak_dt_orswot.erl
```diff
@@ -172,41 +172,25 @@ remove_all([Elem | Rest], Actor, ORSet) ->
...
+
+    {Clock, Entries}.
+
+
 
-%% @private check if either clock dominates the other
--spec either_dominates(riak_dt_vclock:vclock(), riak_dt_vclock:vclock()) ->
-                              riak_dt_vclock:vclock() | concurrent.
-either_dominates(LHSClock, RHSClock) ->
-    case {riak_dt_vclock:descends(LHSClock, RHSClock),
```
lenary, line 201: Is the optimisation worth keeping if you just change these lines to strict_descends? Frankly, this is what I always thought `dominates` meant...

lenary, line 201: Oh, because there is no strict_descends... d'oh.

russelldb, line 201: No, there is a dominates in riak_dt_vclock, but even that is not right. Even if LHS dominates RHS, you still need to compare the contents against RHS's clock.

russelldb:
Gonna YOLO merge as develop is broken, and this is essentially a rollback.

PR #80, opened 2014-02-19
Resurrect Sam & Sean's 'Better' EQC

lenary:
"Better" remains to be seen, but at least intent was clearer, and it tested far more behaviour of
the CRDT itself, which was nice (including serialisation, merges, updates).

It's loltastically out of date. I've done a bit of work when I removed everything related to GC,
but there are still some bugs - indeed the only module i've written callbacks for so far is failing
right now.
- [ ] Use updated `update/3` signature, including precondition fails
- [ ] Write callbacks for all the other modules 
    (some should be put in files in test, ie shared models - ORSet comes to mind)

This is up so @rdb can get eyes on it tomorrow and tell me if I'm an idiot or not.

### DIFF
--- a/src/riak_dt_disable_flag.erl
+++ b/src/riak_dt_disable_flag.erl
@@ -34,6 +34,9 @@
 -ifdef(EQC).
 -include_lib("eqc/include/eqc.hrl").
 -export([gen_op/0, init_state/0, update_expected/3, eqc_state_value/1]).
+
+-behaviour(crdt_statem2_eqc).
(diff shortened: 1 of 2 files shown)


lenary:
So the change to `update/3` makes this change _fuuuun_, because of EQC's symbolic and dynamic values. 

I think I may have to add a `model_update_error_expected(...)` callback or something similar.

seancribbs:
I think @russelldb's new EQC on #83 supersedes any of this work. Let's reopen if/when this needs revisiting.

PR #82, opened 2014-03-07
Fix bugs in Map, Set, and OD Flag

russelldb:
Dear Reviewer,

Let me know if you want me to break this into separate PRs for Map, Set, Flag.

The common theme here is that the merge functions on all three of these types were broken in a similar way. The Set had implementation bugs around common dots. The Flag and Map were broken by design (though the broken design's impl had the same bugs as the Set.) 

---

The Set changes are all about stopping the following bug (or variations on it):

```
add x to a
add x to b
rep a to c
rem x from a
rep b to a
rem x from b
merge (b,merge(a,c))
```

'x' should not be in the set, but the logic used to hold that is 'X' was on both sides of the merge it was in the set. Clearly a value can be on both sides of a merge, but actually be removed, the patch fixes that.

---

The Flag was broken by design. You need more than a vector to track the current value. The result is a lot like the faulty assumption with the set (if both sides are 'true' then the result is 'true') The test that shows this is incorrect is called `disable_test`. The re-write of the Flag is pretty self-explanatory: if there are no `dots` left in the OD Flag's value, it is a `false`, otherwise it is a `true`.

---

The Map problem is rather more complex. As well has having the other bugs above, there is an issue with merging CRDTs during the merge. Looking at the model we used for testing, it only merges CRDTs to a new value on update. It creates a new event that supersedes past events. The problem with our merge function on the Map was that it created this new value, without an event to provoke it. It mixed seen and unseen changes together. Then, when a remove occurred, the values to be removed where bound up with other values and associativity of the merge was broken. Depending on merge order a value could either be in, or out if it got merged in with some data that _was_ concurrent with a remove. The downside to requiring an update event to merge divergent fields, is that you store the field siblings _until_ such an event. So larger maps. I'm yet to quantify exactly how much larger, since the compressed to/from binary seems to take care of _all_ the deduplication. I ran the eqc binary round trip test, hacking it to compare size of the old and new maps, and the new map was usually a little smaller. Go figure. We need to do more testing and benchmarking around this.

I guess I should write a lot more about the new Map design. One thing is for sure, the merge function is a damn sight simpler. And there are still no tombstones.

---

I wrote new EQCs for the Map and Set. They caught a bunch of bugs that old statem test did not. We need to re-write all the CRDTs to use this smaller-state-space style of test. We also need to factor out the commonality between theses tests. I think at this point just raising issues is enough.

Let me know if you need more information. This is last minute, big change stuff, so tread carefully. But be aware, the Map as it is in develop is unshippable, so we need either this design, or something new.

Good luck

Love

RDB


cmeiklejohn:
I can help out on this review on Monday.

seancribbs:
I'm happy to look at it as well, this is critical work.

russelldb:
Yes, I think all the eyes, please.

src/riak_dt_map.erl
```diff
@@ -93,53 +106,25 @@
...
-value(keyset, {_Clock, Values}) ->
-    keys(Values);
-value({contains, Field}, {_Clock, Values}) ->
-    lists:member(Field, keys(Values)).
-
-%% @private
--spec get_crdt({ok, entry()} | error, crdt_mod()) -> crdt() | error.
-get_crdt(Entry, Mod) ->
-    get_crdt(Entry, Mod, false).
+    Remaining = [{Field, CRDT} || {Field, CRDT, _Tag} <- ordsets:to_list(Values)],
```
seancribbs: Do you really need to iterate here to remove the tag? Why not just ignore it in the subsequent fold?

src/riak_dt_map.erl
```diff
@@ -160,94 +145,117 @@ keys(Values) ->
...
 %%
 %% @see riak_dt_orswot for more details.
-
--spec update(map_op(), riak_dt:actor(), map()) -> {ok, map()} | precondition_error().
+-spec update(map_op(), riak_dt:actor() | riak_dt:dot(), map()) ->
+                    {ok, map()} | precondition_error().
+update({update, Ops}, Dot, {Clock, Values}) when is_tuple(Dot) ->
+    NewClock = riak_dt_vclock:merge([[Dot], Clock]),
+    apply_ops(Ops, Dot, NewClock, Values);
+%% Yes, the Map increments the vlcock on field removals!
```
seancribbs: The infamous typo

src/riak_dt_map.erl
russelldb: Good point, left over from when the map had tombstones.

src/riak_dt_map.erl
```diff
@@ -160,94 +145,117 @@ keys(Values) ->
...
-    NewClock = riak_dt_vclock:increment(Actor, Clock),
-    Dot = {Actor, riak_dt_vclock:get_counter(Actor, NewClock)},
-    NewValues = orddict:store(Field, {[Dot], InitialValue}, Values),
-    apply_ops(Rest, Actor, NewClock, NewValues).
+apply_ops([{add, {_Name, Mod}=Field} | Rest], Dot, Clock, Values) ->
+    %% @TODO ¿Should an add read and replace, or stand alone? If it
+    %% stands alone, a concurrent remove results in an empty field,
+    %% which is nice. After an update though, we're back to a mixed,
+    %% unknown result. And of course we have a duplicate entry. Adds
+    %% are weird.
```
seancribbs, line 202: Should we remove "add" from the API? What is the potential negative impact (aside from delaying release)?

src/riak_dt_map.erl
```diff
@@ -160,94 +145,117 @@ keys(Values) ->
...
-              CommonFields).
+merge({LHSClock, LHSEntries}=LHMap, {RHSClock, RHSEntries}=RHMap) ->
+    %% As the clock is incremented for removes as well as adds we can
+    %% optimise here if one clock dominates the other, pick that Map
+    case {riak_dt_vclock:dominates(LHSClock, RHSClock),
+          riak_dt_vclock:dominates(RHSClock, LHSClock)} of
+        {true, false} -> LHMap;
+        {false, true} -> RHMap;
+        _ ->
+            Clock = riak_dt_vclock:merge([LHSClock, RHSClock]),
```
seancribbs, line 219: Can these bits be extracted? This is nested too deep to be readable.

src/riak_dt_map.erl
```diff
@@ -255,50 +263,51 @@ merge_common_fields(CommonFields, Entries1, Entries2) ->
...
 %% @Doc a "fragment" of the Map that can be used for precondition
-%% operations. The schema is just the active Key Set The values are
-%% just those values that are present We use either the values
-%% precondition_context or the whole CRDT
+%% operations. Is the whole map right now :(
 -spec precondition_context(map()) -> map().
 precondition_context(Map) ->
     Map.
 
+%% @Doc stats on internal state of Map.
```
seancribbs: `@doc` should be lowercase.

seancribbs:
- [x] code inspection/style
- [x] eunit and eqc
- [ ] riak_tests
- [x] dialyzer
- [x] documentation

test/orswot_eqc.erl
```diff
@@ -0,0 +1,252 @@
...
+%% Don't replicate to oneself
+replicate_pre(_S, [VN, VN, _]) ->
+    false;
+replicate_pre(_S, [_VN1, _VN2, _]) ->
+    true.
+
+%% Mutating multiple elements in replica_data in place is bad idea
+%% (symbolic vs dynamic state), so instead of treating add/remove and
+%% replicate as the same action, this command handles the replicate
+%% part of the. Data from some random replica (From) is replicated to
```
seancribbs: "the"... what?

test/map_eqc.erl
```diff
@@ -0,0 +1,338 @@
...
+     gen_field(),
+     elements(Replicas), % The replica
+     Cnt,
+     ReplicaData %% The existing vnode data
+    ].
+
+%% Keep the number of possible field names down to a minimum. The
+%% smaller state space makes EQC more likely to find bugs since there
+%% will be more action on the fields. Learned this from
+%% crdt_statem_eqc having to large a state space and missing bugs.
```
seancribbs, line 121: Love this! Good advice for anyone building EQC properties.

src/riak_dt_map.erl
```diff
@@ -255,50 +263,51 @@ merge_common_fields(CommonFields, Entries1, Entries2) ->
...
+stat(duplication, {_, Fields}) ->
+    %% Number of duplicated fields
+    DeDuped = ordsets:fold(fun({Field, _, _}, Acc) ->
+                          ordsets:add_element(Field, Acc) end,
+                  ordsets:new(),
+                  Fields),
+    length(Fields) - length(DeDuped);
 stat(_,_) -> undefined.
 
 -define(TAG, 77).
```
seancribbs: Weren't these moved to a header file?

russelldb: Damn, it never got merged https://github.com/basho/riak_dt/pull/71

seancribbs:
`verify_dt_converge` hangs on check_3a maps. Can you try it and report back?

russelldb:
To be fair, this test hangs on develop without my changes, which suggests some other regression / issue. But yeah, let me look into that.

russelldb:
Actually, cancel that, it only happens with this branch. Oh bum.

seancribbs:
If the new merge function on map can be factored out a bit for clarity and the riak_test fixed, I'm generally in favor of this change and finished reviewing for now. @cmeiklejohn should probably put his eyeballs on it as well.

src/riak_dt_map.erl
russelldb, line 202: I don't know is the only answer I have to give. I guess if we leave it in now, we have it for ever.

russelldb:
So, the riak test is _not_ going to be fixed without some serious work around operations with contexts. At the same time, what does a riak_kv test have to do with the riak_dt library?

My current view is merge this, get it in the next pre, and advise that folk use `include_context=false` until we fix that riak_kv issue.

seancribbs:
Works for me. @cmeiklejohn?

russelldb:
OK, I dealt with all the review comments, and update the riak_test to use `include_context=false` for the time being. I also rebased off develop to get Sam's tag work.

Do you want me to a. force push or b. push to a new branch, open a new PR and reference and close one? (cc @cmeiklejohn again, 'cos why not?!)

russelldb:
See https://github.com/basho/riak_test/pull/550 for riak_test workaround.

seancribbs:
You can force-push to this branch, if you feel comfortable with it.

russelldb:
Done.

seancribbs:
The updated rt is good and works with this branch. You have my :+1:

PR #83, opened 2014-03-19
Context operations

russelldb:
Adds context / deferred operations to riak_dt_od_flag, riak_dt_orswot, riak_dt_map

Uses only a version vector as the context. Comments should help explain it (I hope.)


src/riak_dt_map.erl
```diff
@@ -75,10 +77,14 @@
...
 -type binary_map() :: binary(). %% A binary that from_binary/1 will accept
--type map() :: {riak_dt_vclock:vclock(), entries()}.
+-type map() :: {riak_dt_vclock:vclock(), entries(), deferred()}.
 -type entries() :: ordsets:ordset(entry()).
 -type entry() :: {Field :: field(), CRDT :: crdt(), Tag :: riak_dt:dot()}.
 -type field() :: {Name :: binary(), CRDTModule :: crdt_mod()}.
+%% Only field removals can be deferred. CRDTs stored in the map may
+%% have contexts and deferred operatiosn, but as these are part of the
+%% state, they are stored under the field as an update like any other.
+-type deferred() :: [{context, [field()]}].
```
seancribbs: Did you mean to have `context()` here, not the atom `context`?

src/riak_dt_map.erl
```diff
@@ -99,18 +105,28 @@
                    riak_dt_orswot:orswot_op() | riak_dt_od_flag:od_flag_op() |
                    riak_dt_map:map_op().
 
+-type context() :: riak_dt_vclock:vclock() | undefined.
+
 -type values() :: [value()].
 -type value() :: {field(), riak_dt_map:values() | integer() | [term()] | boolean() | term()}.
 -type precondition_error() :: {error, {precondition, {not_present, field()}}}.
 
+-define(FRESH, riak_dt_vclock:vclock()).
```
seancribbs: This macro seems to be unused.

src/riak_dt_map.erl
```diff
@@ -99,18 +105,28 @@
...
+%% @doc sets the clock in the map to that `Clock'. Used by a
+%% containing Map for sub-CRDTs
+-spec parent_clock(riak_dt_vclock:vclock(), map()) -> map().
+parent_clock(Clock, {_MapClock, Values, Deferred}) ->
+    {Clock, Values, Deferred}.
 
 %% @doc get the current set of values for this Map
 -spec value(map()) -> values().
-value({_Clock, Values}) ->
+value({_Clock, Values, _Deferred}) ->
```
seancribbs, line 127: Why don't deferred operations get applied when considering the value? Are deferred operations only removes in this case?

src/riak_dt_map.erl
```diff
@@ -146,22 +162,39 @@ value(_, Map) ->
...
     NewClock = riak_dt_vclock:increment(Actor, Clock),
     Dot = {Actor, riak_dt_vclock:get_counter(Actor, NewClock)},
-    apply_ops(Ops, Dot, NewClock, Values).
+    {Dot, NewClock}.
 
 %% @private
--spec apply_ops([map_field_update() | map_field_op()], riak_dt:actor(),
-                riak_dt_vclock:vclock(), orddict:orddict()) ->
+-spec apply_ops([map_field_update() | map_field_op()], riak_dt:dot(),
+                {riak_dt_vclock:vclock(), entries() , deferred()}, context()) ->
```
seancribbs, line 191: Does the type here need to be decomposed? Can't it just be `map()`?

src/riak_dt_map.erl
```diff
@@ -173,14 +206,54 @@ apply_ops([{update, {_Name, Type}=Field, Op} | Rest], Dot, Clock, Values) ->
                                         end,
                                         {Type:new(), Values},
                                         FieldInMap),
-    case Type:update(Op, Dot, CRDT) of
+    CRDT1 = Type:parent_clock(Clock, CRDT),
```
seancribbs, line 207: If we expect all nested types to implement this, it should be a callback in the behaviour module.

src/riak_dt_map.erl
lenary, line 127: value() has no way of changing the stored crdt term. 

I believe deferred ops will be carried out if possible at merge-time, which is essentially the only time you get new information to update the internal vvector which would cause deferred operations to be carried out.

russelldb, line 127: As usual @lenary is exactly right. If the contents of `deferred` had any effect on `values` they wouldn't have been deferred. A merge needs to occur before deferred operations' preconditions may be met.

src/riak_dt_map.erl
russelldb, line 191: Dialyzer made me do it, something about an opaque type

src/riak_dt_map.erl
russelldb, line 207: Exactly, I've been moaning on about this in HipChat. Now I have to retrofit all the CRDTs with `update/4` and `parent_clock`. I din't do it yet, hence not in the behaviour.

seancribbs, line 207: Yeah, the other option is to do the module_info trick and see if it is exported.

src/riak_dt_map.erl
russelldb: yeah

src/riak_dt_map.erl
russelldb: yeah, an aborted idea I had. Forgot to clean up

src/riak_dt_map.erl
russelldb, line 127: And yes, only field removals are deferred.

src/riak_dt_map.erl
russelldb, line 207: Right. Not sure which is best. I think I prefer it in the behaviour. But I'm not sure which is best.

seancribbs, line 207: Behaviour is fine with me, lmk if you want me to handle that drudgery.

lenary:
I made a test for ORSWOT, that i thought this code wouldn't do what a user might expect. Turns out it passes (and the user expectation is what's done), _yay_: https://gist.github.com/lenary/84f68e865f6a54e9832f

src/riak_dt_map.erl
```diff
@@ -173,14 +206,54 @@ apply_ops([{update, {_Name, Type}=Field, Op} | Rest], Dot, Clock, Values) ->
...
+    ToAdd = {Field, Mod:new(), Dot},
+    NewValues = ordsets:add_element(ToAdd, Values),
+    apply_ops(Rest, Dot, {Clock, NewValues, Deferred}, Ctx).
+
+%% @private depending on context, call type's update/3 or update/4 fun
+-spec update_field(crdt_mod(), crdt_op(), riak_dt:dot(), crdt(), context()) ->
+                          {ok, crdt()} | precondition_error().
+update_field(Type, Op, Dot, CRDT, undefined) ->
+    Type:update(Op, Dot, CRDT);
+update_field(Type, Op, Dot, CRDT, Ctx) ->
```
seancribbs: Is it possible for the nested type not to care about context? If so, does it even make sense to call `update/4`? Why not always call `update/4` and pass `undefined` as the last argument?

src/riak_dt_map.erl
```diff
@@ -192,36 +265,61 @@ apply_ops([{remove, Field} | Rest], Dot, Clock, Values) ->
...
-    %% stands alone, a concurrent remove results in an empty field,
-    %% which is nice. After an update though, we're back to a mixed,
-    %% unknown result. And of course we have a duplicate entry. Adds
-    %% are weird.
-
-    ToAdd = {Field, Mod:new(), Dot},
-    NewValues = ordsets:add_element(ToAdd, Values),
-    apply_ops(Rest, Dot, Clock, NewValues).
+%% Context removes
+remove_field(Field, {Clock, Values, Deferred0}, Ctx) ->
```
seancribbs, line 261: Just to be clear: this clause defers the removal (as long as the context is not an ancestor), but still prunes elements that are dominated by the removal clock?

seancribbs, line 261: Sorry, your edoc comment answered my question, the answer is "yes".

src/riak_dt_map.erl
```diff
@@ -264,12 +362,36 @@ keep_or_drop(Clock, Tag, Entries, Field) ->
             ordsets:add_element(Field, Entries)
     end.
 
+
+%% @private apply those deferred field removals, if they're
+%% preconditions have been met, that is.
+-spec apply_deferred(riak_dt_vclock:vclock(), entries(), deferred()) ->
+                            {riak_dt_vclock:vclock(), entries(), deferred()}.
+apply_deferred(Clock, Entries, Deferred) ->
```
seancribbs, line 360: You are now applying the removals, but it seems that when going through the `remove_field/3` function, these removals can either be deferred again, or dropped because their context is dominated by the map's clock. Am I misunderstanding this?

seancribbs, line 360: Ah, but perhaps you've already covered that because any fields whose dots are dominated by the context clock will still be removed.

src/riak_dt_od_flag.erl
```diff
@@ -81,23 +123,41 @@ merge({LHSClock, LHSDots}, {RHSClock, RHSDots}) ->
...
-    %% Perfectly possible that an item in both sets should be dropped
-    {NewClock, Flag}.
+
+    Deferred = ordsets:union(LHSDeferred, RHSDeferred),
+
+    apply_deferred(NewClock, Flag, Deferred).
+
+apply_deferred(Clock, Flag, Deferred) ->
+    lists:foldl(fun(Ctx, ODFlag) ->
+                        {ok, ODFlag2} = update(disable, a, ODFlag, Ctx),
```
seancribbs: Does passing an arbitrary actor here make sense?

src/riak_dt_orswot.erl
```diff
@@ -143,12 +153,80 @@ update({add_all, Elems}, Actor, ORSet) ->
...
 %% none are.
 update({remove_all, Elems}, Actor, ORSet) ->
     remove_all(Elems, Actor, ORSet).
 
+-spec update(orswot_op(), actor() | dot(), orswot(), riak_dt:context()) ->
+                    {ok, orswot()} | precondition_error().
+update({add, Elem}, Actor, ORSet, _Ctx) ->
+    ORSet = add_elem(Actor, ORSet, Elem),
+    {ok, ORSet};
+update({remove, Elem}, _Actor, {Clock, Entries, Deferred}, Ctx) ->
```
seancribbs, line 168: Brilliant. Just brilliant.

src/riak_dt_orswot.erl
```diff
@@ -246,7 +355,7 @@ merge_common_keys(CommonKeys, {LHSClock, LHSEntries}, {RHSClock, RHSEntries}) ->
               CommonKeys).
 
 -spec equal(orswot(), orswot()) -> boolean().
-equal({Clock1, Entries1}, {Clock2, Entries2}) ->
+equal({Clock1, Entries1, _}, {Clock2, Entries2, _}) ->
```
seancribbs, line 360: Interesting thought... if their clocks are equal, should they not be equal regardless?

src/riak_dt_orswot.erl
```diff
@@ -281,37 +390,47 @@ update_entry(Elem, Entries, Dot) ->
...
+%%  an operation is needed at a replica without sending the entire
+%%  state to the client. In the case of the ORSWOT the context is a
+%%  version vector. When passed as an argument to `update/4' the
+%%  context ensures that only seen adds are removed, and that removes
+%%  of unseen adds can be deferred until they're seen.
 -spec precondition_context(orswot()) -> orswot().
-precondition_context(ORSet) ->
-    ORSet.
+precondition_context({Clock, _Entries, _Deferred}) ->
+    %% @TODO? ¿If we merge all the deferred contexts in, then the target of
```
seancribbs: Is this comment still relevant? I don't understand the question.

test/map_eqc.erl
```diff
@@ -176,6 +180,53 @@ remove_next(S=#state{replica_data=ReplicaData, adds=Adds}, Res, [Add, _]) ->
...
+    case choose_field(FromMap) of
+        empty ->
+            {From, FromMap, FromModel};
+        {Ctx, Field} ->
+            {ok, Map} = riak_dt_map:update({update, [{remove, Field}]}, To, ToMap, Ctx),
+            Model = model_ctx_remove(Field, FromModel, ToModel),
+            {To, Map, Model}
+    end.
+
+choose_field(Map) ->
```
seancribbs, line 213: This feels a little suspect, could you generate an integer in the `_args` callback that can be shrunk?

test/map_eqc.erl
```diff
@@ -312,19 +363,48 @@ model_update_field({Name, Type}=Field, Op, Actor, Cnt, {Adds, Removes}=Model) ->
...
+    Removes = sets:union(Removes1, Removes2),
+    Deferred = sets:union(Deferred1, Deferred2),
+    model_apply_deferred(Adds, Removes, Deferred).
+
+model_apply_deferred(Adds, Removes, Deferred) ->
+    D2 = sets:subtract(Deferred, Adds),
+    ToRem = sets:subtract(Deferred, D2),
+    {Adds, sets:union(ToRem, Removes), D2}.
+
+    %% sets:fold(fun(E, {A, R, D}) ->
```
seancribbs: What does this comment mean? Dead code?

test/map_eqc.erl
```diff
@@ -312,19 +363,48 @@ model_update_field({Name, Type}=Field, Op, Actor, Cnt, {Adds, Removes}=Model) ->
...
+
+model_merge({Adds1, Removes1, Deferred1}, {Adds2, Removes2, Deferred2}) ->
+    Adds = sets:union(Adds1, Adds2),
+    Removes = sets:union(Removes1, Removes2),
+    Deferred = sets:union(Deferred1, Deferred2),
+    model_apply_deferred(Adds, Removes, Deferred).
+
+model_apply_deferred(Adds, Removes, Deferred) ->
+    D2 = sets:subtract(Deferred, Adds),
+    ToRem = sets:subtract(Deferred, D2),
```
seancribbs, line 383: Is this statement not just `sets:intersection(Deferred, Adds)`?

test/map_eqc.erl
```diff
@@ -312,19 +363,48 @@ model_update_field({Name, Type}=Field, Op, Actor, Cnt, {Adds, Removes}=Model) ->
...
+    %%           {Adds, Removes, sets:new()},
+    %%           Deferred).
+
+model_ctx_remove(Field, {FromAdds, _FromRemoves, _FromDeferred}, {ToAdds, ToRemoves, ToDeferred}) ->
+    %% get adds for Field, any adds for field in ToAdds that are in
+    %% FromAdds should be removed any others, put in deferred
+    ToRemove = sets:filter(fun({F, _Val, _Token}) -> F == Field end, FromAdds),
+    %% [{F, Val, Token} || {F, Val, Token} <- sets:to_list(FromAdds), F == Field],
+    Defer = sets:subtract(ToRemove, ToAdds),
+    Remove = sets:subtract(ToRemove, Defer),
```
seancribbs, line 392: Again double-subtract vs. intersection

src/riak_dt_map.erl
seancribbs: Also note that in the coverage report I don't see this clause being called at all. Shouldn't the `map_eqc` at least call it once?

seancribbs:
Generally I'm in-favor of this change. It greatly improves the semantics of map and orswot and reduces the size of the context considerably. There are a few nitpicks and holes that I have commented on. It's hard to ensure that I've covered everything, but the EQC for both map and orswot are really good now and I think I understood most every addition except for the parts that I have commented on.

I'm happy to do a final review as soon as you're ready.

src/riak_dt_orswot.erl
lenary, line 360: Yeah, they should be equal.

They might not be if from a base ORSWOT, the actor `a` added one element, and from the same base ORSWOT added a distinct element, making the vclocks be the same, but the actual entries different, but that can't happen in Riak (unless something very odd is going on)

src/riak_dt_orswot.erl
lenary: Yeah, I don't understand it either. I don't understand how we "fast-forward" the clock,, given we never merge with the context- vclock, we just check if the current vclock supersedes it.

src/riak_dt_od_flag.erl
lenary: It doesn't matter, the `disable` update ignores the actor provided.

src/riak_dt_map.erl
russelldb: The map_eqc does not do this yet, no. There are no context updates of fields 'cos the model has no context. I'm thinking of adding a version vector to the model for this.

As for your first comment, I don't know, maybe `update/4` should handle `undefined`. I'll look at it today.

src/riak_dt_map.erl
russelldb, line 360: Yes. Basically, if the Map clock does not descend the context, there may be an add we haven't seen that we still need to remove when we've seen it. So the op has to be deferred still, though we can at least partially execute it on dots the context has seen.

src/riak_dt_od_flag.erl
russelldb: Yeah, but it might be cleaner to _do_ the disable rather than delegate to update/4. I'll see if I can refactor.

src/riak_dt_orswot.erl
russelldb, line 360: I think it can happen in riak, due to kv#679. I'd rather assume the worst here.

src/riak_dt_orswot.erl
russelldb: Yeah, my thinking on it is hazy, I'll cut the comment out.

test/map_eqc.erl
russelldb, line 213: I don't know. It would involve knowing how many fields are in the map, so you'd need it's concrete, not symbolic state.

test/map_eqc.erl
russelldb, line 383: oh yeah, derp.

russelldb, line 383: Except, how do I get the new deferred? Still need to subtract, at that point, we're looking at two statements anyway, so what is the difference, except preference?

test/map_eqc.erl
russelldb, line 392: See above comment. I'm happy with this, do you think it makes it harder to understand?

src/riak_dt_orswot.erl
seancribbs, line 360: Noted.

test/map_eqc.erl
seancribbs, line 213: Perhaps it's not worth it, I just worry about shrinking being affected.

test/map_eqc.erl
seancribbs, line 383: Yeah, I suppose it doesn't make a difference.

test/map_eqc.erl
russelldb, line 213: I just don't know how, do you?

seancribbs, line 213: Does the model have enough information that one could generate the offset/index from that?

russelldb, line 213: I don't understand, the model when concrete does, but not some symbolic value.

seancribbs:
Assigned 2.0-beta milestone.

seancribbs:
+1 5aae7c4

russelldb:
Is that +1 to the PR, or just that commit?

seancribbs:
@russelldb Can't it be bolth?

russelldb:
Heh, why not?! Ok, here goes, all 17 commits!

PR #85, opened 2014-04-14
Make sure all data types use configured compression in t2b/b2t

lenary:
For 2.0 release (map and orswot currently use compression):
- [x] lwwreg (only ever in map, which never calls to_binary/1 in lwwreg)
- [x] od-flag (only ever in map, which never calls to_binary/1 in od-flag)
- [x] pn-counter (we've kinda looked at this, but compression might do better)

For later:
- [ ] gcounter
- [ ] gset
- [ ] oe-flag
- [ ] orset

### DIFF
--- a/src/riak_dt.erl
+++ b/src/riak_dt.erl
@@ -22,6 +22,7 @@
 
 -module(riak_dt).
 
+-export([to_binary/1, from_binary/1]).
 -export_type([actor/0, dot/0, crdt/0, context/0]).
 
 -type crdt() :: term().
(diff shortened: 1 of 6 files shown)


seancribbs:
`riak_dt_orswot` uses the compression setting incorrectly. We should wrap the setting in a centralized function so as to reduce copy-pasta.

lenary:
Going to attack this now, because bored in a class.

lenary:
lww-reg and pn-counter will be incompatible with current representations if we use the centralised functions. 

What's the status of breaking changes right now?

russelldb:
Status is "do it!" https://www.youtube.com/watch?v=M9-IZ1pNk00

lenary:
Actually, after a closer reading, not going to be incompatible for pncounter V2 serialisation (which is the one we aim to upgrade to anyway).

seancribbs:
:+1: 786008f

seancribbs:
@borshop merge

PR #91, opened 2014-04-15
Use testing_time in favor of numtests in eunit tests

Vagabond:
This helps the tests avoid timing out on slower (read buildbot)
machines.


seancribbs:
:+1: 929a510

seancribbs:
@borshop merge

PR #93, opened 2014-05-14
Reset-remove semantic on Map

russelldb:
With a shared causal context for the map and the CRDTs it embeds, removing a field, and then re-adding it has the effect of removing all the additions to that field when merging with a concurrent update to the field. For example:

```
Add 'X' to set in field F at A
Merge with B
Add 'Y' to set in field F at B
remove field F at A
Add 'Z' to field F at A
merge A and B
```

The set contains only 'Y' and 'Z', since the remove of the field at A containing 'X' constitutes a remove of 'X'.

However, if the field was _not_ re-added by A the semantic was different, and the result would contain 'X, Y, Z'. This seemed counter-intuitive and confusing. These changes add a temporary tombstone to present fields, so that the removal of a field effects a removal of all dots seen by the field at removal time. But you don't have to re-add the field to see this effect. Neat.

One intuitive way to think of it as like an offline sync of a file system like Dropbox. With this semantic, removing a directory on one machine, and adding a file to that directory on another, merges to the directory being present, with only the added file in it.

There is some oddness with counters, though.

I added a new counter type, `riak_dt_emcntr` (see it's docs for more details) that stores a `dot` as latest event against each actors increments and decrements. This is so a field removal can reset the counter value. However, we don't store a dot per increment (that would be crazy inefficient). Counters work will in this scenario

```
A increments counter at field F by 10
B merges A
A removes field F
B increments counter at field F by 5
A and B merge
```

Counter is at 5 (A's dot for the increment is dropped)

However there is an anomaly

```
A increments counter at field F by 10
B merges A
A increments counter by 5
B removes field F
A and B merge
```

The counter is at 15.

Now, technically, A's first increment _should_ be dropped (should it??) and only the second stand, but that does not happen, as we don't keep a dot per increment. This, too, can be spun in the filesystem sense. If you updated a file and concurrently removed it, you wouldn't expect only the diff to survive, but the whole file.

Personally I'm happy enough with the semantic, to enable reset-remove on counters requires a dot per increment (a counter as set of increments) and that is not worth the overhead.

Happy to rebase once the full round of reviewing is done.

OOOOOOOH, And I removed `add` from the Map API. Problem?

### DIFF
--- a/src/riak_dt_emcntr.erl
+++ b/src/riak_dt_emcntr.erl
@@ -128,9 +128,11 @@ op(increment, {P, N}) ->
     {P+1, N};
 op(decrement, {P, N}) ->
     {P, N+1};
-op({increment, By}, {P, N}) ->
+op({_Op, 0}, {P, N}) ->
+    {P, N};
+op({increment, By}, {P, N}) when is_integer(By), By > 0 ->
     {P+By, N};
-op({decrement, By}, {P, N}) ->
+op({decrement, By}, {P, N}) when is_integer(By), By > 0 ->
     {P, N+By}.
 
 %% @doc takes two `emcntr()'s and merges them into a single
@@ -261,8 +263,7 @@ gen_op() ->
     oneof([increment,
            {increment, nat()},
            decrement,
-           {decrement, nat()},
-           {increment, ?LET(X, nat(), -X)}
+           {decrement, nat()}
           ]).
 
 update_expected(_ID, increment, Prev) ->
--- a/src/riak_dt_emcntr.erl
+++ b/src/riak_dt_emcntr.erl
@@ -1,0 +1,390 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_dt_emcntr: A convergent, replicated, state based PN counter,
+%% for embedding in riak_dt_map.
+%%
+%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc A PN-Counter CRDT. A PN-Counter is essentially two G-Counters:
+%% one for increments and one for decrements. The value of the counter
+%% is the difference between the value of the Positive G-Counter and
+%% the value of the Negative G-Counter. However, this PN-Counter is
+%% for using embedded in a riak_dt_map. The problem with an embedded
+%% pn-counter is when the field is removed and added again. PN-Counter
+%% merge takes the max of P and N as the merged value. In the case
+%% that a field was removed and re-added P and N maybe be _lower_ than
+%% their removed values, and when merged with a replica that has not
+%% seen the remove, the remove is lost. This counter adds some
+%% causality by storing a `dot` with P and N. Merge takes the max
+%% event for each actor, so newer values win over old ones. The rest
+%% of the mechanics are the same.
+%%
+%% @see riak_kv_gcounter.erl
+%%
+%% @reference Marc Shapiro, Nuno Preguiça, Carlos Baquero, Marek Zawirski (2011) A comprehensive study of
+%% Convergent and Commutative Replicated Data Types. http://hal.upmc.fr/inria-00555588/
+%%
+%% @end
+
+-module(riak_dt_emcntr).
+-behaviour(riak_dt).
+
+-export([new/0, value/1, value/2]).
+-export([update/3, merge/2, equal/2]).
+-export([to_binary/1, from_binary/1]).
+-export([stats/1, stat/2]).
+-export([parent_clock/2, update/4]).
+
+%% EQC API
+-ifdef(EQC).
+-include_lib("eqc/include/eqc.hrl").
+-export([gen_op/0, update_expected/3, eqc_state_value/1, init_state/0, generate/0]).
+-endif.
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-type emcntr() :: {riak_dt_vclock:vclock(), [entry()]}.
+-type entry() :: {Actor :: riak_dt:actor(), {Event :: pos_integer(),
+                                             Inc :: pos_integer(),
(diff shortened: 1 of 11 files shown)


russelldb:
This is kind of WIP, in that there is still W to do. But I thought it prudent to get eyes on early.

Still todo:-
- [x] docs for map (see todo in `riak_dt_map`)
- [x] riak_test for semantic
- [x] remove copypasta test from `riak_dt_emcntr`
- [x] riak_kv support of `riak_dt_emcntr`
- [ ]  riak_kv removal of map `add` (or re-add `add` to map with better semantic)

src/riak_dt_map.erl
```diff
@@ -293,16 +339,118 @@ defer_remove(Clock, Ctx, Field, Deferred) ->
                                 Deferred)
     end.
 
-%% @doc merge two `map()'s.  and then a pairwise merge on all values
-%% in the value list.  This is the LUB function.
+%% @doc merge two `map()'s.
 -spec merge(map(), map()) -> map().
+merge(Map, Map) ->
+    Map;
+%% @TODO is there a way to optimise this, based on clocks maybe?
```
seancribbs, line 444: You could short-circuit the case where one clock dominates the other, returning only the dominant one. Right?

russelldb, line 444: That requires more analytical thinking than I'm capable of right now (thinking embedded maps more than the container.) Maybe…?

seancribbs, line 444: Oh, right. I'll ponder it a while too, maybe we can figure it out tomorrow.

seancribbs, line 444: Well, tomorrow came and went and we didn't come to any conclusion. That said, I can't see why it isn't safe to ignore one map entirely if its clock is dominated (or even just descended). My intuition says that these no-op merges would be mostly at the get_fsm where replicas are already in-sync, and so there would still be value there, by way of reduced latency.

seancribbs:
Can you rebase or merge develop into this so it will merge cleanly (and make borshop happy)?

russelldb:
After the review I can.

src/riak_dt_emcntr.erl
```diff
@@ -0,0 +1,396 @@
...
+                {P::non_neg_integer(), N::non_neg_integer()}.
+op(Op, error) ->
+    op(Op, {0, 0});
+op(Op, {ok, {_Evt, P, N}}) ->
+    op(Op, {P, N});
+op(increment, {P, N}) ->
+    {P+1, N};
+op(decrement, {P, N}) ->
+    {P, N+1};
+op({increment, By}, {P, N}) ->
```
seancribbs: Nitpick: This doesn't handle the subtly-evil user who increments or decrements by a negative amount.

src/riak_dt_emcntr.erl
```diff
@@ -0,0 +1,396 @@
...
+
+init_state() ->
+    0.
+
+gen_op() ->
+    oneof([increment,
+           {increment, nat()},
+           decrement,
+           {decrement, nat()},
+           {increment, ?LET(X, nat(), -X)}
```
seancribbs: Well, ignore my previous comment about `op/2`, it looks like negative increments are handled in the quickcheck property.

seancribbs: But, I still don't see how negative increments are handled properly. This doesn't have the EQC test across it by itself.

src/riak_dt_map.erl
```diff
@@ -70,37 +84,46 @@
...
 
--type crdt()  ::  riak_dt_pncounter:pncounter() | riak_dt_od_flag:od_flag() |
+-type crdt()  ::  riak_dt_emcntr:emcntr() | riak_dt_od_flag:od_flag() |
                   riak_dt_lwwreg:lwwreg() |
                   riak_dt_orswot:orswot() |
                   riak_dt_map:map().
 
 -type map_op() :: {update, [map_field_update() | map_field_op()]}.
 
--type map_field_op() :: {add, field()} | {remove, field()}.
```
seancribbs, line 101: Reminder to remove this op from `riak_pb` and `riak_kv_crdt_json`.

src/riak_dt_map.erl
```diff
@@ -244,42 +267,65 @@ apply_ops([{add, {_Name, Mod}=Field} | Rest], Dot, {Clock, Values, Deferred}, Ct
...
+ctx_rem_field({_, Type}, {ok, {CRDTs, TS0}}, Ctx, MapClock) ->
+    %% Drop dominated fields, and update the tombstone.
+    %%
+    %% If the context is removing a field at dot {a, 1} and the
+    %% current field is {a, 2}, the tombstone ensures that all events
+    %% from {a, 1} are removed from the crdt value. If the ctx remove
+    %% is at {a, 3} and the current field is at {a, 2} then we need to
+    %% remove only events upto {a, 2}. The glb clock enables that.
+    %%
+    %% @TODO document how this doesn't really work for counters.
```
seancribbs: Does the narrative in the PR description cover this TODO?

seancribbs:
Is it ok that the `map_eqc` tends to have `0` as the most common actor count? Ran it for 6 minutes and got this output:

```
19.94% 0
11.44% 4
8.55% 3
8.31% 2
8.15% 6
[...]
```

src/riak_dt_map.erl
russelldb: Yes, does that count as "document"? If so, will happily cut that `@TODO` from src

russelldb:
@seancribbs good question, I need to look at that some more. I mean, that's a high percentage, I think I need further metrics to see what size the map is, 'cos I suspect that there is still an issue with the state for this test. Luckily I have some expert eyes tomorrow.

src/riak_dt_map.erl
seancribbs: Yeah, maybe we just need a separate TODO for internal and external documentation.

src/riak_dt_map.erl
seancribbs, line 444: @russelldb On second thought, could deferred removes happen without incrementing the clock? If so, then my suggestion is moot.

russelldb, line 444: Going to let eqc tell me. Deferreds are applied in merge, which merges the clocks.

russelldb, line 444: OK, deferring this TODO for another branch. Want to get this semantic into the next Beta, optimisation comes after.

seancribbs:
@russelldb I've rebased your branch, cleaning up a duplicate commit and grouping similar work. To get your local copy working:
1. `git stash` any unsaved changes you have
2. `git fetch origin` to get the latest head of the branch
3. `git reset --hard origin/feature-bug/rdb/removes-remove` to move your local branch to the rebased one
4. `git stash apply` to reapply your local changes

Ping me if you have trouble with this.

russelldb:
OK. Addressed review comments and pushed. Just going to remove `add` from kv api, then ready for final review.

russelldb:
Removed "add" from PB and HTTP APIs https://github.com/basho/riak_pb/pull/93, https://github.com/basho/riak_kv/pull/951

seancribbs:
:+1: ffce004

seancribbs:
@borshop merge

PR #94, opened 2014-06-09
Optimise dot metadata, add creates a new dot, discards old

russelldb:
As in the Map, add should drop all previous dots and replace with a new frontier dot.


lenary:
+1 bb3d0b8

russelldb:
@borshop merge

PR #95, opened 2014-06-14
Do not sort lists of ops given to OR-Sets

russelldb:
Sorting means that batch operations applied by the same actor
behave differently to separate operations applied linearly.
There is no need to sort, it is a throwback from some mistaken
assumptions in early development, before we had context removes.


lenary:
+1 6d63176

russelldb:
@borshop merge

PR #97, opened 2014-06-19
Support negative counter increments and decrements

lenary:
Apparently a boog in emcounter. Fixed, so server-side should have no errors with any integers and any of `increment` and `decrement`.

### DIFF
--- a/src/riak_dt_emcntr.erl
+++ b/src/riak_dt_emcntr.erl
@@ -70,8 +70,9 @@
                  }.
 
 
--type emcntr_op() :: riak_dt_gcounter:gcounter_op() | decrement_op().
--type decrement_op() :: decrement | {decrement, pos_integer()}.
+-type emcntr_op() :: increment_op() | decrement_op().
+-type increment_op() :: increment | {increment, integer()}.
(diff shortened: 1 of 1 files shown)


russelldb:
How do I test this? Please provide the steps that provoke the bug this fixes, so I can verify. Also, a _test_ that shows the bug and the fix would be very valuable (be it a riak_test, or a local unit test)

macintux:
I don't think this is entirely correct: IIRC from looking at this code yesterday the Erlang specs indicate positive integers only, so that will have to be tweaked (although buildbot didn't gripe about Dialyzer).

I also suspected this should be handled one layer higher in the stack, but haven't looked to see where that would be.

Testing should be easy: fire up the Erlang client and send a single decrement. That should crash a node IIRC.

Will look more in a bit.

Thanks @lenary, really appreciate you getting this started.

lenary:
@russelldb there's a new eunit test in 9109d7e

russelldb:
Thanks @lenary this all works fine now

+1 9109d7e

macintux:
I assume we need to fix the specs before merging

macintux:
@borshop merge

PR #102, opened 2014-08-03
Add LWW-Element-Set with Add-Wins semantic for equal timestamp

russelldb:
An LWW-Element-Set similar to Roshi/The Shapiro Tech Report. When an add and remove on the same element have the same timestamp, the add wins. Why? Makes sense to me that you only remove a element you've seen in the set, and if you remove `a` at time `1` then the `a` you saw must have been added at some time `<1`. Therefore an add of `a` at `1` is not the same `a` the remove wants to remove.

If there is a need for a remove wins version, we could tweak this impl to provide either based on some setup param, I guess.

Despite what the Roshi docs say, there is no _GC_, and there are in fact tombstones. Once an element is added it will always take up space in the datastructure.

Also, "doomstones" allowed (you can remove `{a, 100000}` when `a` is not present, and no add of `a` less than `100000` will work.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## WHAT?
 
-A set of state based CRDTs implemented in Erlang and  on the paper -
+A set of state-based CRDTs implemented in Erlang, and based on the paper -
 [A Comprehensive study of Convergent and Commutative Replicated Data Types]
 (http://hal.inria.fr/docs/00/55/55/88/PDF/techreport.pdf) - which you
 may find an interesting read.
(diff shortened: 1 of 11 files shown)


cmeiklejohn:
Two dialyzer errors:

```
riak_dt_lwwset.erl:109: Invalid type specification for function riak_dt_lwwset:add_elem/3. The success typing is (_,_,[{_,_}]) -> [{_,_}]
riak_dt_lwwset.erl:123: Invalid type specification for function riak_dt_lwwset:remove_elem/3. The success typing is (_,_,[{_,_}]) -> [{_,_}]
```

peterbourgon:
> Despite what the Roshi docs say, there is no GC, and there are in fact tombstones.

I'll update Roshi's docs to clarify my (perhaps eccentric) use of the term GC, and explicitly mention the possibility of doomstones (hehe). Thanks for writing!

russelldb:
Re-based on develop. Brought up to date.

`make test` passes (including eqc)
dialyzer passes

Just needs a friendly lil' +1 from someone out there…

PR #106, opened 2014-10-03
orswot_eqc.erl causes riak_dt tests to appear to fail

russelldb:
If you don't have EQC installed, orswot_eqc won't compile, which
looks bad. If def out that eqc stuff.


seancribbs:
:+1: 270c7eb

PR #110, opened 2014-12-04
Swap orddict for dict in orswot and map

russelldb:
After a report on IRC of Maps with 13k set elements inside taking up to 80seconds to merge, I ran some benchmarks comparing `dict` and `orddict` (as I had a suspicion this was the bulk of problem.)

This PR swaps orddict for dict. There may well be many more optimisations to do, but as the attached graph, and tables below show, this is a mighty big speed up.

This plot is insert times for dict vs orddict in an orswot.
insert times into a set

Merge times, orswot, fully probably\* fully disjoint sets with probably*\* 20 actors orddict vs dict (times are microseconds)

[_Probably fully disjoint as crypto:rand_bytes(30) was used to generate elements, they could clash I guess. *_Probably 20 actors as actor chosen at random from a list of 20 randomly generated 8 byte actor ids, but if you have fewer than 20 elements, unlikely you have 20 actors, eh?]

| Elements | orddict | dict |
| --- | --- | --- |
| 100000 | 1455538575 | 745421 |
| 10000 | 9621000 | 39468 |
| 1000 | 99874 | 3286 |
| 100 | 1467 | 428 |
| 10 | 54 | 64 |
| 1 | 11 | 24 |

---

This plot compares merge times for the code that used lots of simple set operations, vs. an uglier, single fold over the unioned keyspace of the map.

merge5_times


lenary:
I thought this would bite us. Someone has a branch with hashdict stuff (from elixir). Sean will know about it, though maybe it was my branch. I haven't seen graphs/tables, so I don't know the comparison, but worth checking.

russelldb:
Those branches will be stale. I asked about HashDict, not ready for primetime. Hopefully this will do for now.

Wish you'd said earlier that you thought this was an issue.

russelldb:
Still todo
- [x] add `to/from_binary/2` to the `riak_dt` behaviour
- [x] impl `to/from_binary/2` for other data types

include/riak_dt_tags.hrl
```diff
@@ -35,3 +35,8 @@
 -define(DT_MAP_TAG, 77).
 
 %% Other
+
+%% Error macros
+-define(UNSUPPORTED_VERSION(V), {error, unsupported_version, Vers}).
```
seancribbs: Vers will not work here, you need to make the argument to the macro common with the variable in the tuple.

src/riak_dt_map.erl
```diff
@@ -204,7 +205,10 @@
 %% Only field removals can be deferred. CRDTs stored in the map may
 %% have contexts and deferred operations, but as these are part of the
 %% state, they are stored under the field as an update like any other.
--type deferred() :: [{context(), [field()]}].
+-type deferred() :: dict(context(), [field()]).
+
+%% used until we move to erlang 17 and can use dict:dict/2
+-type dict(_A, _B) :: dict().
```
seancribbs, line 211: Might be good to wrap this in an ifdef so that we can provide 17-compatible typespecs as well. Copy-pasta from some other repo that does it should be sufficient.

src/riak_dt_map.erl
```diff
@@ -660,16 +669,58 @@ stat(_,_) -> undefined.
...
 -spec to_binary(map()) -> binary_map().
 to_binary(Map) ->
-    <<?TAG:8/integer, ?V1_VERS:8/integer, (riak_dt:to_binary(Map))/binary>>.
+    {ok, B} = to_binary(?V1_VERS, Map),
+    B.
+
+%% @private encode v1 maps as v2, and vice versa. The first argument
+%% is the target binary type.
+-spec to_binary(Vers :: pos_integer(), map()) -> {ok, binary_map()} | ?UNSUPPORTED_VERSION.
+to_binary(?V1_VERS, Map0) ->
```
seancribbs, line 678: Do we have a `to_binary/2` for version 2? And a `?V2_VERS` macro? Or are we just downgrading to v1 on serialization for some reason?

src/riak_dt_map.erl
```diff
@@ -845,6 +896,24 @@ equals_test() ->
     ?assert(equal(C, D)),
     ?assert(equal(A, A)).
 
+%% %% up/downgrade tests
+%% v1_v2_test() ->
```
seancribbs: Can this test be reinstated?

russelldb: I guess, except this PR no longer changes the map binary version.

seancribbs: Aha, I see now, the at-rest representation is the same. Thanks.

src/riak_dt_map.erl
russelldb, line 678: isn't one anymore (though there will be if the new map format stuff gets in) went with a common at rest format to simplify the change (no KV component, no capabilities etc.)

seancribbs, line 678: Aha, yes. Thanks!

seancribbs:
:+1: 44383b1

seancribbs:
@borshop merge

Issue #99, opened 2014-07-21
Deferred operations may be dropped before they're delivered


russelldb:
# Deferred Ops

Distributed databases like basho/riak sometimes use sloppy quorums for greater availability. Because of this I added `deferred operations` to data types that have a "observed remove" semantic.

In the CRDT Catalogue tech report[1] removing an element that is not present from a set (and likewise a field from a map) results in a "precondition exception". It is a precondition of removal that the element be present.

In riak, Data types are a mixed operation/state based hybrid. The client sends operations to the database, and the database performs these operations on the state of the data type.

We provide a `precondition_context`  or `context` for short when a data type is read, and this is returned with any remove operation to enforce the "observed remove" semantic.

Due to sloppy quorums (called fallbacks in Riak) it is possible that a replicas receiving an operation may lag behind the causal history observed by a client. In the extreme, a fallback my start up just to perform the operation and have _no_ state whatsoever.

Deferred operations were added to mitigate this scenario. If a fallback receives a context remove for a set element that is not present, it stores the context, and operation, as part of the data type state. When later, the set merges with other replicas, the operation will be performed as soon as the Set's causal history is up-to-date with the deferred operations.
## Maps

There is a bug with deferred operations when a Set/Flag/Map is stored in a Map. Here is the scenario. All examples are R/W=1 for simplicity, but any disjoint set of R/W will do.
- Client reads Map `[{set_field, [a, b, c]}]` from replica X
- Client sends op `remove a from set_field` with X context
- Empty replica Y handles op (creates Map, adds `set_field` with deferred op)
- Client reads Map from Y (gets `[{set_field, []}]` as result (field is present, no values, just deferred operation)
- Client sends `remove set_field` to Y with Y context
- The deferred operation is lost when the field is removed on Y

The fix, I think, is either to store deferred operations at the top-level, or never remove a field a with undelivered operations (though do not show such a field in `value`.) Whichever is simpler/works best.

seancribbs:
@russelldb Is this likely to be fixed in the 2.0 cycle? Otherwise I'll remove the known-issue label.

PR #113, opened 2015-02-03
Valter's work addressing basho/riak_dt#99 

russelldb:
I've opened this pull request to keep track of comments/feedback as I work through the code.


balegas:
The current version of the map clears deferred operations when an entry is removed (and no dot survives), allowing some dots that should be deleted by a deferred operation to remain in the state: 
- OpA removed "b -> 1" in replica A with the context of B;
- A's context does not descend from B, deferring OpA; 
- OpA1 removes entry b, clears the entry and the deferred op in replica A; 
- A and B merge, "b -> 1" is added back to A and not cleared by OpA as intended because the deferred operation was cleared by OpA1

In a nutshell, i've changed the Map data-type to avoid losing these deferred operations. 
I added some form of tombstone in the state that tells that an entry is removed, but keeps the value while it has deferred operations that are not covered by the object's clock. 
Changes were made in the remove operation to fill this tombstone when an entry with deferred operations is deleted.
Merge() was modified to clear tombstones
Value() was changed to hide elements that are removed but the tombstone is not covered by the map clock yet.

The remove operation now calls a propagate_remove() function that recursively checks if any child of the current map entry has any deferred operations. All deferred operations are merged into the object clock and propagated to the root element that was deleted, filling the tombstone of the map entries on the way.
While an element has a tombstone, the entry and its subtree are kept (immediately remove dots that are covered by the clock). When the merge occurs and the deferred operations arrive, clear_tombstones() checks what entries can be deleted: if any child received a new dot, that entry is kept; If no new dot was added the entry can be deleted (also checks empty entry) --- the tombstone is only erased when covered by the map clock. To read the object, similar rules are applied: if an entry has a tombstone, check if any child has dots that are not covered by the tombstone to show/hide the entry.

Since the amount of meta-data in the object state was growing, i decided to separate the meta-data from the actual value so a map entry is now {{Dots, Seen, Tombstone},CRDT}.

PR #115, opened 2015-02-21
Add support for Erlang 17.

cmeiklejohn:
Add support for Erlang 17 by using namedspaced types.  Also, change the
map() type to riak_dt_map(), given there's a name collision with the
built-in map and as of Erlang 18, it will no longer be allowed to
redefine types.


seancribbs:
:+1: 24839d3

This is basically what I did in my throwaway attempt. Good work!

cmeiklejohn:
@borshop merge

PR #121, opened 2016-03-24
doc related fix & explanation

liuzhen:
1. lwwreg is actually not DVV based for this implementation.
2. an extra explanation of the not-so-straightforward-acronym orswot would be of help to non-native speakers.


russelldb:
+1 9a7f855dbfddb8087d2ea5e716c8bb120ba202f4

russelldb:
Many thanks for the speedy response. Gratefully received.

PR #125, opened 2016-07-22
Enable gset support

binarytemple-external:
Re-targeting GSET support onto develop-2.2 branch


binarytemple-external:
@russelldb - what do you think ?

PR #133, opened 2017-09-18
Fix `equal/2` bug around unordered dict usage

russelldb:
When `riak_dt_map` moved from `orddict` to `dict` the `equal/2` function
was not updated to reflect that the `deferred` operations list was now
an unordered `dict`, meaning two equal maps could report as unequal just
because of the order of operations in the deferred dict. The bug was
found by the added counter-example
`test/non-commute-counter-example.eqc`.

I also added a function `map_eqc:check/1` so that any counter-example
file can be easily run from the shell.

Run `./rebar eunit skip_deps=true deps_dir=../ compile_only=true` and
then `erl -pa .eunit/` and
`map_eqc:check("test/non-commute-counter-example.eqc")` before and after
the change to `riak_dt_map.erl` to verify the fix.


russelldb:
Taking plus one on #135 as a plus one here

PR #132, opened 2017-04-26
Fix dialyzer

lucafavatella:

PR #137, opened 2019-10-03
Mas 3.0 update

martinsumner:
There are two parts tot his PR.  First the gset code which was missing from develop-3.0 is re-added in.

The second change is restricted to the dialyzer specs in `riak_dt_map`.  This was previously leading to dialyzer failures.  Looking at the code there were types:

field_name() :: {binary(), crdt_mod()}
field() :: {field_name(), field_value()}

There were many function specs that expected field(), but looking at the actual functions they appeared to want field_name() instead.  So I removed the definition of field(), and used field_name() throughout - and that pleased dialyzer.  However, I'm not clear on map nomenclature - should this be called field() and not field_name()?  Is this more than a labelling issue, have I misunderstood the problem?

@russelldb appreciate you might not have the time, but if you do have the chance let me know if I've done the right thing with regards to the `riak_dt_map` specs for field()/field_name().


src/riak_dt_lwwreg.erl
```diff
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% riak_dt_lwwreg: A DVVSet based last-write-wins register
+%% riak_dt_lwwreg: A last-write-wins register
 %%
 %% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
```
ThomasArts, line 5: you keep the 2013 here, whereas you moved to 2016 in riak_dt_gset?

martinsumner: This was all done as part of the gset implementation, so no idea why one was changed but not the other.  I don't think there is any consistency over copyright notices on changes throughout the repos.  I don't know what the right answer is if define a new module, or modify an existing one.  Something to discuss with @martincox

ThomasArts (approved):

martinsumner:
The field_name()/field() switch was also in another open PR - https://github.com/basho/riak_dt/pull/132/files

